PR #14, opened 2023-10-24
KIL-2578: [Agent] Redshift proxy client

mrostan:
- New Redshift proxy client that uses `psycopg2` library.
- Created a base class: `BaseDbProxyClient` that should work in the future as the base class for other database clients like Postgres, MSSQL, etc.
- Added code to discard cached clients if an exception was raised during the execution of an operation, noticed that Redshift clients keep raising errors after an Exception occurred.


delkopiso (approved):
looks good

PR #19, opened 2023-10-27
KIL-2629: Logging improvements

mrostan:
- Provides a way to enable debug logging by setting env var `MCD_DEBUG_LOG` to `true`
- Includes a debug message when an operation completes, including the elapsed time
- Updates the names of the attributes in log messages used for trace_id and operation_name to `mcd_trace_id` and `mcd_operation_name`
- Uses `logging.Filter` feature to update the logging context including `trace_id` and `operation_name` for all requests, this ends up including this information in all logs, including those from other libraries, helping to get all logs related to a single operation when filtering by `mcd_trace_id`.


delkopiso (approved):
looks good

PR #23, opened 2023-11-14
KIL-2644 Add azure storage implementation

delkopiso:
Moves the Azure Storage client implementation from [data colletor](https://github.com/monte-carlo-data/data-collector/blob/eeeaa4679217fcc94de73817d941bdcff3ce0508/lambdas/clients/azure_blob/reader_writer.py) to agent.


mrostan (approved):
lgtm!

PR #24, opened 2023-11-14
Added endpoint to get outbound ip address

mrostan:


pxc-dev (approved):
Cool stuff! Not required, but as a nice to have it might be useful if we can include this as an optional part of the health check endpoint too (e.g. via a flag that is disabled by default)

mrostan:
Good idea @pxc-dev , added a "full" parameter to the health endpoint, if present "extra" information like the outbound ip address will be included

PR #25, opened 2023-11-14
KIL-2644 Relax dependency requirements

delkopiso:
References the `requirements.in` file in `setup.py` instead of `requirements.txt` to make the project more easily consumable as a library.


mrostan (approved):
lgtm, thanks!

PR #26, opened 2023-11-15
KIL-2723 Postgres proxy client

delkopiso:
#### What's new?
- New Postgres proxy client that uses psycopg2 library.

### DIFF
--- a/apollo/integrations/db/postgres_proxy_client.py
+++ b/apollo/integrations/db/postgres_proxy_client.py
@@ -1,0 +1,45 @@
+from typing import Dict, Optional
+
+import psycopg2
+from psycopg2 import DatabaseError
+from psycopg2.errors import QueryCanceled, InsufficientPrivilege  # noqa
+
+from apollo.integrations.db.base_db_proxy_client import BaseDbProxyClient
+
+_ATTR_CONNECT_ARGS = "connect_args"
+
+
+class PostgresProxyClient(BaseDbProxyClient):
+    """
+    Proxy client for Postgres.
+    Credentials are expected to be supplied under "connect_args" and will be passed directly to `psycopg2.connect`, so
+    only attributes supported as parameters by `psycopg2.connect` should be passed.
+    If "autocommit" is present in credentials it will be set in _connection.autocommit.
+    """
+
+    def __init__(self, credentials: Optional[Dict], **kwargs):  # type: ignore
+        if not credentials or _ATTR_CONNECT_ARGS not in credentials:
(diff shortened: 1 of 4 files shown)


apollo/integrations/db/postgres_proxy_client.py
```diff
@@ -0,0 +1,45 @@
...
+
+_ATTR_CONNECT_ARGS = "connect_args"
+
+
+class PostgresProxyClient(BaseDbProxyClient):
+    """
+    Proxy client for Postgres.
+    Credentials are expected to be supplied under "connect_args" and will be passed directly to `psycopg2.connect`, so
+    only attributes supported as parameters by `psycopg2.connect` should be passed.
+    If "autocommit" is present in credentials it will be set in _connection.autocommit.
```
mrostan, line 17: It looks like this comment about `autocommit` doesn't apply for postgres

apollo/integrations/db/postgres_proxy_client.py
```diff
@@ -0,0 +1,45 @@
...
+import psycopg2
+from psycopg2 import DatabaseError
+from psycopg2.errors import QueryCanceled, InsufficientPrivilege  # noqa
+
+from apollo.integrations.db.base_db_proxy_client import BaseDbProxyClient
+
+_ATTR_CONNECT_ARGS = "connect_args"
+
+
+class PostgresProxyClient(BaseDbProxyClient):
```
mrostan, line 12: Just wondering, should we have this class extending the Redshift one? or maybe the other way around? It looks the code is almost the same

tests/test_postgres_client.py
```diff
@@ -0,0 +1,189 @@
...
+    ATTRIBUTE_NAME_RESULT,
+    ATTRIBUTE_NAME_ERROR_TYPE,
+)
+from apollo.agent.logging_utils import LoggingUtils
+
+_POSTGRES_CREDENTIALS = {
+    "host": "www.test.com",
+    "user": "u",
+    "password": "p",
+    "port": "5439",
```
mrostan, line 19: it's not relevant, but I guess this should be 5432 just to be consistent with PG default values

mrostan (approved):
lgtm! left a question/comment about re-using/sharing the code with Redshift client

apollo/integrations/db/postgres_proxy_client.py
delkopiso: I think redshift extending postgres makes sense as Redshift uses Postgres under the hood

PR #30, opened 2023-11-20
KIL-2744 Fix invalid execute params type for SQL Server proxy client

delkopiso:
- Uses a custom cursor class `SqlServerProxyClientCursor` to override the query parameters and provided them as a tuple instead of a list for SQL Server's client
- Fixes the [sentry error](https://monte-carlo-data.sentry.io/issues/4643357426/?alert_rule_id=11714577&alert_type=issue&environment=dev&notification_uuid=4153e0de-851e-4d12-895c-3bed8cfdcf0f&project=3527635&referrer=slack): `'params' arg (<class 'list'>) can be only a tuple or a dictionary.`

### DIFF
--- a/apollo/integrations/db/sql_server_proxy_client.py
+++ b/apollo/integrations/db/sql_server_proxy_client.py
@@ -1,6 +1,7 @@
 from typing import (
     Any,
     Dict,
+    Iterable,
     Optional,
 )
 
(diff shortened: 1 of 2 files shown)


apollo/integrations/db/sql_server_proxy_client.py
```diff
@@ -28,3 +51,6 @@ def __init__(self, credentials: Optional[Dict], **kwargs: Any):
     @property
     def wrapped_client(self):
         return self._connection
+
+    def cursor(self) -> Any:
+        return SqlServerProxyClientCursor(self.wrapped_client.cursor())
```
mrostan, line 56: let's add a comment explaining why we're doing this

mrostan (approved):
it looks great, left a comment just to add a comment explaining the reason for doing this

PR #31, opened 2023-11-20
KIL-2729 Snowflake Proxy Client

mrostan:
- New Snowflake proxy client
- Added support for `error_extra_attributes` used to return `errno` and `sqlstate` for Snowflake exceptions, so we can create the same exceptions client side with that information


delkopiso (approved):
Nice!

PR #33, opened 2023-11-21
KIL-2744 Oracle DB proxy client

delkopiso:
#### What's new?
- Adds a new `OracleProxyClient` for processing agent executions with connection type `oracle`.
- Adds a `_process_description` hook to the `BaseDbProxyClient` that can be overridden to customize the serialization of the returned description data.
- Overrides `_process_description` in `OracleProxyClient` to handle the presence of the `DbType` values Oracle returns in place of the column type values.


mrostan (approved):
lgtm!

PR #36, opened 2023-11-22
KIL-2770 Support responding with pre-signed URLs

delkopiso:
#### What's new?
- Calculates the size, in bytes, of the agent's response and if `response_size_limit_bytes` is set in the agent operation object, then the result is written to the configured storage and the written object's location is returned as a pre-signed URL instead.
- Adds a new `AgentSerializer` JSON encoder that uses the pre-existing serialization logic

### DIFF
--- a/apollo/agent/agent.py
+++ b/apollo/agent/agent.py
@@ -381,19 +381,20 @@ class Agent:
             ),
         )
         response = AgentResponse(result or {}, 200, operation.trace_id)
-        size = response.calculate_result_size()
-        if operation.use_pre_signed_url(size):
-            key = f"responses/{operation.trace_id}"
-            storage_client = StorageProxyClient(self._platform)
-            storage_client.write(key=key, obj_to_write=response.serialize_result())
-            expiration_seconds = int(
-                os.getenv(
-                    PRE_SIGNED_URL_RESPONSE_EXPIRATION_SECONDS_ENV_VAR,
-                    PRE_SIGNED_URL_RESPONSE_EXPIRATION_SECONDS_DEFAULT_VALUE,
+        if operation.can_use_pre_signed_url():
+            size = response.calculate_result_size()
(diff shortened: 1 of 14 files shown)


tests/test_gcs_storage_proxy.py
delkopiso, line 1: needed to tweak these tests because the size calculation was throwing `AssertionError: 'Object of type MagicMock is not JSON serializable' is not None`

apollo/agent/agent.py
```diff
@@ -374,7 +380,21 @@ def _execute_client_operation(
                 dict(elapsed_time=time.time() - start_time),
             ),
         )
-        return AgentResponse(result or {}, 200, operation.trace_id)
+        response = AgentResponse(result or {}, 200, operation.trace_id)
+        size = response.calculate_result_size()
```
mrostan, line 384: It looks like pre-signed urls are disabled when `operation.response_size_limit_bytes` is 0, so maybe we should check here before calling `calculate_result_size` as it seems a little bit expensive

apollo/agent/serde.py
```diff
@@ -0,0 +1,35 @@
...
+        elif isinstance(value, date):
+            return {
+                ATTRIBUTE_NAME_TYPE: ATTRIBUTE_VALUE_TYPE_DATE,
+                ATTRIBUTE_NAME_DATA: value.isoformat(),
+            }
+        return value
+
+    def default(self, obj: Any):
+        serialized = self.serialize(obj)
+        if serialized is not obj:  # no serialization happened
```
mrostan, line 33: shouldn't the comment here be " # serialization happened" ?

mrostan (approved):
looks good to me!, left a few minor comments/questions

apollo/agent/serde.py
delkopiso: yup good catch

PR #42, opened 2023-11-30
Fix git clone error in Lambda agent

mrostan:
- In order to install `oscrypto` from GitHub ([PR](https://github.com/monte-carlo-data/apollo-agent/pull/40) we had to install git (`yum install git -y`)
- This left `git` installed in the Docker image used by the Lambda function which breaks the ability to use `git` in that image. In Lambda, `git` needs to be used through the `lambda-git` library (that installs a git version that works in this environment) and not the one you can install in the OS.
- This PR creates a separate stage in the Dockerfile that installs git and downloads pip dependencies to install `oscrypto` from GitHub, the final stage copies the packages from the previous one without installing git, so we have the required libraries without having `git` installed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,7 +52,7 @@ RUN apt install git -y
 
 CMD . $VENV_DIR/bin/activate && gunicorn --timeout 930 --bind :$PORT apollo.interfaces.cloudrun.main:app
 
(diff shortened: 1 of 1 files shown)


delkopiso (approved):
nice!

PR #44, opened 2023-11-30
Ssmith/kil 2804 add agent support for azure dedicated sql pool

sassmith:
Add agent support for Azure Dedicated SQL pool

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,9 +21,12 @@ RUN . $VENV_DIR/bin/activate && pip install --no-cache-dir -r requirements.txt
 RUN . $VENV_DIR/bin/activate && pip install setuptools==65.5.1
 
 # Azure Dedicated SQL Pools uses pyodbc which requires unixODBC and 'ODBC Driver 17 for SQL Server'
-RUN apt-get update && \
-    apt-get install -y unixodbc unixodbc-dev && \
-    ACCEPT_EULA=Y apt-get install -y msodbcsql17
+RUN apt-get update
+    && apt-get install -y curl apt-transport-https \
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -20,6 +20,14 @@ RUN . $VENV_DIR/bin/activate && pip install --no-cache-dir -r requirements.txt
...
 RUN . $VENV_DIR/bin/activate && pip install setuptools==65.5.1
 
+# Azure Dedicated SQL Pools uses pyodbc which requires unixODBC and 'ODBC Driver 17 for SQL Server'
+RUN apt-get update \
+    && apt-get install -y gnupg gnupg2 gnupg1 curl apt-transport-https \
+    && curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add - \
+    && curl https://packages.microsoft.com/config/debian/10/prod.list > /etc/apt/sources.list.d/mssql-release.list \
+    && apt-get update \
+    && ACCEPT_EULA=Y apt-get install -y msodbcsql17 unixodbc unixodbc-dev
+
```
mrostan, line 30: you need to install the same packages using `yum` in the `lambda` image at the end of this same `Dockerfile` and test this with a lambda agent

mrostan (commented):
it looks good Sam, left a comment about supporting the Lambda agent

Dockerfile
sassmith: added and then tested with test-lambda-agent account's non-vpc collector and it worked
Screenshot 2023-12-01 at 2 21 27 PM

mrostan (approved):
nice!

PR #47, opened 2023-12-05
New endpoints to retrieve logs for Lambda and CloudRun agents

mrostan:
- New platform specific endpoints for `aws` and `gcp`:
- GCP:
  - `/api/v1/gcp/logs/list`: gets CloudRun logs using `google.cloud.logging_v2.client.Client.list_entries`
    - By default, it returns all logs for the last 10 minutes
    - It supports the following parameters:
      - `limit`: defaults to `1000`
      - `filter`: a filter expression, see: [docs](https://cloud.google.com/logging/docs/view/advanced_filters)
- AWS:
  - `/api/v1/aws/logs/filter`: gets `CloudWatch` logs using `CloudWatchLogs.filter_log_events`, docs [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/logs/client/filter_log_events.html).
    - By default, it returns all logs for the last 10 minutes
    - It supports the following parameters:
      - `limit`: defaults to `1000`
      - `start_time`: start time in isoformat (defaults to `now - 10 minutes`)
      - `end_time`: end time in isoformat
      - `pattern`: a search pattern, docs [here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/FilterAndPatternSyntax.html)
  - `/api/v1/aws/logs/start_query`: starts a query using `CloudWatchLogs.Client.start_query`, docs [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/logs/client/start_query.html).
    - By default, it returns all logs matching the required `query` in the last 10 minutes
    - It supports the following parameters:
      - `limit`: defaults to `1000`
      - `start_time`: start time in isoformat (defaults to `now - 10 minutes`)
      - `end_time`: end time in isoformat
      - `query`: a query, syntax docs [here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CWL_QuerySyntax.html).
    - It returns a `query_id` that can be used with `query_results`
  - `/api/v1/aws/logs/query_results`: returns the log events for the specified query id
    - The only supported parameter is `query_id`


Dockerfile
```diff
@@ -97,4 +92,9 @@ RUN yum -y update \
 RUN curl https://packages.microsoft.com/config/rhel/7/prod.repo | tee /etc/yum.repos.d/mssql-release.repo
 RUN ACCEPT_EULA=Y yum install -y msodbcsql17
 
+COPY apollo "${LAMBDA_TASK_ROOT}/apollo"
+ARG code_version="local"
+ARG build_number="0"
+RUN echo $code_version,$build_number > ./apollo/agent/version
+
```
mrostan, line 99: Moved this as the last step, so when building locally and making code changes it builds the updated image faster without having to re-install the yum packages

delkopiso (approved):
Looks good! Just a couple of questions:
- We have a default limit of 1000 for the logs. Should we have a maximum limit? Or are we just relying on the platform to impose their own limits here?
- For CloudWatch logs, do we want to support `stopQuery`?

PR #49, opened 2023-12-05
pins base docker image version to older version to keep support for yum

sassmith:
### What's New
With this new AWS release to their Docker images, `yum` is being replaced with `dnf`.
@mrostan and I's testing unfortunately proved this will not be a straight forward switch

This PR pins the base docker version to before this release so that `yum` is still supported. Will investigate switching over to `dnf` in a future ticket


mrostan (approved):
looks good, thanks!

PR #51, opened 2023-12-08
KIL-2846 Tableau proxy client

delkopiso:
#### What's new?
- Adds a Tableau proxy client for making requests to the Tableau API via both the tableauserverclient SDK's metadata query API and HTTP requests to Tableau REST API.
- Each tableau operation performs a sign in using a newly generated JWT. This makes it unnecessary to issue sign in operations to the agent.
- Each JWT is generated with a configurable expiration time. This expiration time has a default value of 5 minutes, but can be set by the caller when initializing the client similar to how other database clients handle statement timeouts.


mrostan (approved):
looks good!

PR #53, opened 2023-12-13
KIL-2852: Lambda Direct Update for AWS Agent

mrostan:
- New `LambdaDirectUpdater` Agent Updater:
  - Enabled when the wrapper type (from `MCD_AGENT_WRAPPER_TYPE` env var) is not `CLOUDFORMATION`.
  -  Uses [update_function_code](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda/client/update_function_code.html) to update the image used by the Lambda Function.
  - Supports updating `MemorySize` and `ConcurrentExecutions` by specifying these parameters in the update operation.
  - It uses [update_function_configuration](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda/client/update_function_configuration.html) to update the memory size.
  - It uses [put_function_concurrency](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/lambda/client/put_function_concurrency.html) to update the concurrency settings.
- AWS Platform Provider refactored as it was named after "CloudFormation" but it actually works even in CloudFormation is not used, for example to access `CloudWatch` logs.


delkopiso (approved):
Nice!

PR #55, opened 2023-12-15
add sap hana proxy client

sassmith:
Adds proxy client for SAP HANA and unit tests

Tested in `test.aws_agent` with the AWS remote agent, was able to create integration and run SQL monitor


mrostan (approved):
nice!

PR #58, opened 2023-12-19
KIL-2848 PowerBI proxy client

delkopiso:
#### What's new?
- Adds a new PowerBI proxy client that wraps the existing HTTP Proxy client in order to generate an access token by performing an Azure Active Directory token exchange.


mrostan (approved):
nice!

PR #59, opened 2023-12-21
KIL-2910: update HTTP proxy client to support custom auth headers

97nitt:
This PR updates the `HttpProxyClient` to support custom authentication headers. 

For example, [authentication for the Pinecone API](https://docs.pinecone.io/docs/authentication#add-a-header-to-an-http-request) requires an API token be passed in an `Api-Key` header.


mrostan (approved):
looks good!

PR #60, opened 2023-12-22
KIL-2929 Support setting SSL verification on HTTP proxy client

delkopiso:
#### What's new?
Allows passing an optional boolean `verify_ssl` argument to the HTTP proxy client. If present, this value controls the `verify` argument to the `requests.request` method which determines whether the server's TLS certificate is verified or not.


mrostan (approved):
looks good!

PR #61, opened 2023-12-22
KIL-2919: [Azure Agent] implement get_logs endpoint

mrostan:
- Azure logging improvements to log structured data as we log for the other platforms
- Added Flask error handler to log unhandled exceptions
- Added `/api/v1/azure/logs/query` endpoint to get Azure logs using `Azure Monitor Query` client library
- The role `Log Analytics Reader` is required to be assigned to the user running the Azure Function at the Application Insights resource


apollo/interfaces/azure/host.json
```diff
@@ -3,8 +3,7 @@
   "logging": {
     "applicationInsights": {
       "samplingSettings": {
-        "isEnabled": true,
-        "excludedTypes": "Request"
+        "isEnabled": false
```
mrostan, line 6: disabled sampling as we want all messages to be logged

delkopiso (approved):
Looks great!

PR #62, opened 2023-12-26
add support for execute_sql_query with signed URL response

santiagoaguiar:
To support running dynamic and sample queries from the frontend, we want to return a signed URL from the agent so that the UI can download the file directly.

As we can have no further processing on the data that what the agent is able to perform, we need to define a common format and transform to that format in the agent (instead of doing the transformation on the data collector).

This PR adds an implementation for Snowflake to validate the approach and open it for suggestions & comments.


apollo/agent/models.py
```diff
@@ -88,5 +94,15 @@ class AgentHealthInformation:
         metadata=config(exclude=exclude_none_values), default=None
     )
 
-    def to_dict(self) -> Dict:  # type: ignore
-        pass
+
+@dataclass
+class AgentExecuteSqlQueryResponse(DataClassJsonMixin):
```
santiagoaguiar, line 99: Opted to declare this here as a 'built-in command response schema', so it could be easier to adhere to in the different clients.

apollo/agent/serde.py
```diff
@@ -48,3 +55,27 @@ def default(self, obj: Any):
         if serialized is not obj:  # serialization happened
             return serialized
         return super().default(obj)
+
+
+def row_value_encoder(value: Any) -> Any:
```
santiagoaguiar, line 60: Encoding is different here than the default agent serializer, we want it to be more lightweight, and typing to be inferred if needed from field schema (which we still need to add...).

apollo/integrations/snowflake/snowflake_proxy_client.py
```diff
@@ -48,3 +50,27 @@ def get_error_extra_attributes(self, error: Exception) -> Optional[Dict]:
...
+            else:
+                is_partial = False
+
+            description = cursor.description or []
+
+            return AgentExecuteSqlQueryResponse(
+                number_of_rows_fetched=len(results),
+                field_names=[field[0] for field in description],
+                rows=results,
+                is_partial=is_partial,
```
santiagoaguiar, line 75: Might change this schema a little. Now, it mimics what the data collector is returning on the execute SQL query, so we don't need a different implementation depending if we are going against an agent or a legacy DC.

apollo/agent/agent.py
```diff
@@ -457,6 +462,14 @@ def _execute_client_operation(
                 )
                 url = storage_client.generate_presigned_url(key, expiration_seconds)
                 response.use_location(url)
+                logger.info(
+                    f"Generated pre-signed url for operation: {connection_type}/{operation_name}",
```
santiagoaguiar, line 466: Logging here as generating pre-signed URL seems significant enough, as it expands the lifetime of the returned data and makes it available to others.

apollo/interfaces/agent_response.py
```diff
@@ -56,5 +56,10 @@ def calculate_result_size(self) -> int:
             return 0
         return len(self.serialize_result().encode())
 
-    def serialize_result(self) -> str:
-        return json.dumps(self.result, cls=AgentSerializer)
+    def serialize_result(self, unwrap_result: bool = False) -> str:
```
santiagoaguiar, line 59: I need the data to be stored without any agent message wrapping, as the data is intended to be used by clients directly.

mrostan (approved):
looks good to me, nice work!

PR #63, opened 2023-12-28
Add TCP Keep Alive settings for Redshift and Postgres connections

mrostan:
- In Azure we were experiencing issues due to missing/wrong keep alive settings connecting to Redshift, basically queries taking more than 5 minutes were never ending
- The issue was fixed by adding the following settings when the `psycopg` connection is initialized:
  - `keepalives=1`
  - `keepalives_idle=30`
  - `keepalives_interval=10`
  - `keepalives_count=5`
- More information about these settings [here](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-KEEPALIVES)
- **Note**: we'll need to test all other integrations as they might experience the same issue. It would be great if we could change the settings globally for the image, as it is possible in Linux as discussed [here](https://stackoverflow.com/questions/71343175/why-are-the-keepalives-params-in-psycopg2-connect-required-to-run-long) and [here](https://github.com/moby/moby/issues/31208), but it seems it's not possible to change these settings for a Docker image.


apollo/integrations/db/postgres_proxy_client.py
```diff
@@ -22,9 +23,18 @@ def __init__(self, credentials: Optional[Dict], client_type: str = "postgres", *
...
-        )
+        }
+        # we were having tcp keep alive issues in Azure, so we're forcing it now unless configured from the dc
+        # https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-KEEPALIVES
+        if _KEEP_ALIVE_SETTING not in connect_args:
+            connect_args[_KEEP_ALIVE_SETTING] = 1
+            connect_args[f"{_KEEP_ALIVE_SETTING}_idle"] = 30
+            connect_args[f"{_KEEP_ALIVE_SETTING}_interval"] = 10
+            connect_args[f"{_KEEP_ALIVE_SETTING}_count"] = 5
+
```
santiagoaguiar, line 36: Very small nit, but maybe just easier to do keepalives_idle (easier to search later for these settings if needed :)).

santiagoaguiar (approved):

apollo/integrations/db/postgres_proxy_client.py
mrostan: Yeah, good point, will make the change

PR #65, opened 2023-12-29
KIL-2884: update readme file with instructions for Lambda and Azure

mrostan:
Updated Readme file that was focused on GCP development only to support all platforms and link when needed to our public docs.
Left an "advanced" section with some commands needed for manual deployment.


README.md
```diff
@@ -58,7 +57,34 @@ deployment options.
 }
 ```
 
-### Adding new integrations
+#### Running storage operations locally
+If you need to run storage (S3, GCS or Azure Blob) operations locally you can run the agent setting these 
+environment variables: `MCD_STORAGE` and `MCD_STORAGE_BUCKET_NAME`, for example:
+```shell
+PYTHONPATH=. MCD_DEBUG=true MCD_STORAGE_BUCKET_NAME=agent-bucket MCD_STORAGE=GCS python apollo/interfaces/generic/main.py
```
santiagoaguiar, line 64: Not a bad idea to include a `.envrc.example` file or similar on the repo with all the env vars that you usually need to set, though they are not that many today.

BTW, instead of `PYTHONPATH=.` it also works to run `python -m apollo.interfaces.generic.main`

README.md
```diff
@@ -58,7 +57,34 @@ deployment options.
...
+            {
+                "method": "list_objects"
+            }
+        ]
+    }
+}
+```
+
+## Adding new integrations
+How to add new integrations is documented in Notion [here](https://www.notion.so/montecarlodata/Adding-support-for-Remote-Agents-to-integrations-3d5025ef36eb47de8488cdafdc39d42c?pvs=4).
```
santiagoaguiar, line 87: Given this is a public repo, we probably shouldn't be adding notion links here. Maybe that doc could be moved to the repo.

README.md
```diff
@@ -76,29 +102,28 @@ You can use `BqProxyClient` as a reference, basically you just need to:
...
-gcloud run deploy dev-apollo-agent --image montecarlodata/pre-release-agent:latest-cloudrun --region us-east1
-```
-The same build process publishes also the `generic` image with the tag `latest-generic`, so you can use it to run a
-standalone agent.
 
-### Release process
+For now, that build is not updating the dev agents for the different platforms, the easiest way to update 
+them is to connect to MC Dev environment with `dev.apollo.agent` user (credentials in 1Pwd), go to
+Settings -> Integrations -> Agents & Data Store and update the (or use the `agents upgrade` command in MC CLI).
+
```
santiagoaguiar, line 112: Similar about these comments, we might need to move them elsewhere?

santiagoaguiar (approved):

README.md
pxc-dev: You can just link to our public docs for the process
- AWS: https://docs.getmontecarlo.com/docs/create-and-register-an-aws-agent#how-do-i-upgrade-the-agent
- GCP: https://docs.getmontecarlo.com/docs/create-and-register-a-gcp-agent#how-do-i-upgrade-the-agent

README.md
```diff
@@ -117,3 +159,55 @@ See [LICENSE](https://github.com/monte-carlo-data/apollo-agent/blob/main/LICENSE
 ## Security
 
 See [SECURITY](https://github.com/monte-carlo-data/apollo-agent/blob/main/SECURITY.md) for more information.
+
+## Advanced deployment
```
pxc-dev, line 163: Let's include a comment making it clear that the preferred way to deploy is using our templates and that these templates also support using any image, so devs can just use a test image with them. 

Using IaC is easier and more manageable than creating resources manually.

pxc-dev: The examples in the template repos might also be handy to have links for too as they automate usage

E.g. with AWS Terraform -- https://github.com/monte-carlo-data/terraform-aws-mcd-agent/blob/70c571c5bf0e698f3b640fef8ffd03cc3e8372e1/examples/agent/Makefile#L10

README.md
```diff
@@ -117,3 +159,55 @@ See [LICENSE](https://github.com/monte-carlo-data/apollo-agent/blob/main/LICENSE
...
+
+If you have your own ECR repo used for testing you'll need to login first:
+```shell
+aws ecr get-login-password --region us-east-1 --profile <aws_profile> | docker login --username AWS --password-stdin <account_id>.dkr.ecr.us-east-1.amazonaws.com
+```
+
+And then tag/push your image:
+```shell
+docker tag lambda_agent:latest <account_id>.dkr.ecr.us-east-1.amazonaws.com/dev-agent:95
+docker push <account_id>.dkr.ecr.us-east-1.amazonaws.com/dev-agent:95
```
pxc-dev, line 206: These steps hardcode version numbers. We should make it clear that those numbers are just examples

README.md
```diff
@@ -58,7 +61,45 @@ deployment options.
...
-### Adding new integrations
+#### Local execution using Docker
+You can also execute the agent building and running a Docker image:
+```shell
+docker build -t local_agent --target generic --platform=linux/amd64 .;\
+docker run --rm --name local_agent -p8081:8081 -ePORT=8081 -it local_agent
+```
+Or running the latest dev image from DockerHub:
+```shell
+docker run --rm --name dev_agent -p8081:8081 -ePORT=8081 -it montecarlodata/pre-release-agent:latest-generic
```
pxc-dev, line 72: For future devs, can we include a line explaining how to run the unit tests with the image? Thanks!

pxc-dev (approved):
Thanks again for doing this!

PR #67, opened 2024-01-03
[Azure Agent] Cleanup Durable Functions history periodically

mrostan:
- Azure Durable Functions store data for all requests (input and output data) forever, and as discussed [here](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-instance-management?tabs=python#purge-instance-history) we can use the `Orchestration Client API` to manually purge the history for a given instance or all instances created in a given period.
- This PR introduces a timer-triggered function that executes every 12 hours (hard-coded at 3 AM and 3 PM) and deletes the history for all completed (failed or succeeded) instances that were executed more than 1 day ago.

### DIFF
--- a/apollo/interfaces/azure/function_app.py
+++ b/apollo/interfaces/azure/function_app.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+from datetime import datetime, timedelta
 from typing import Dict
 
 from azure.monitor.opentelemetry import configure_azure_monitor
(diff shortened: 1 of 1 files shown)


apollo/interfaces/azure/function_app.py
```diff
@@ -143,3 +144,37 @@ def agent_api(req: func.HttpRequest, context: func.Context):
...
+    schedule="0 0 3,15 * * *", arg_name="timer", run_on_startup=False
+)  # run every day at 3 AM/PM
+@app.durable_client_input(client_name="client")
+async def cleanup_durable_functions_data(
+    timer: func.TimerRequest, client: DurableOrchestrationClient
+) -> None:
+    created_time_from = datetime.today() - timedelta(
+        days=365 * 10
+    )  # datetime.min or None not supported
+    created_time_to = datetime.today() - timedelta(days=1)
```
delkopiso, line 160: should these times be timezone-aware? if the cron-schedule is UTC based, then we'll probably want these to be UTC as well.

delkopiso (approved):
looks good, just one question

apollo/interfaces/azure/function_app.py
mrostan: Good call, I think it's works fine with the local tz but I agree it's better to be timezone aware, made the changed and monitored the next execution, everything looks good. Thanks

PR #69, opened 2024-01-05
KIL-2957 Athena proxy client

delkopiso:
#### What's new?
- Adds a new `AthenaProxyClient` that extends `BaseAwsProxyClient` for operations with `athena` connection type.
- Overrides the `process_result` method to fix the incorrect serialization of query timestamps.


mrostan (approved):
looks good

PR #70, opened 2024-01-09
Enabled TCP Keep Alive for MySQL connections

mrostan:
MySQL connections in Azure agents were having the same issue fixed for Redshift with PR https://github.com/monte-carlo-data/apollo-agent/pull/63 
This PR adds code to force keep-alive to be enabled for MySQL connections and to send keep alive packets every 30 seconds.


97nitt (approved):
LGTM!

PR #71, opened 2024-01-11
KIL-2988: Add support for compressing large results

mrostan:
Two new settings are now supported for operations:
- `compress_response_file`: controls whether file saved to the bucket (when a pre-signed url is returned) is compressed or not, defaults to `False`.
- `compress_response_threshold_bytes`: when the size is larger than this threshold and a pre-signed url is not used, the response is compressed and the result is sent with `Content-Type: application/gzip`. Defaults to `0` (disabled).


delkopiso (approved):
looks great!

PR #72, opened 2024-01-11
KIL-3031 VULN-230 Upgrade pip, urllib3, jinja2 and pycryptodome

mrostan:
- This PR fixes: "Medium vulnerabilities identified in the packages cryptography, urllib, ncurses*, python, libcurl, and pip"
- The native packages (ncurses* and libcurl) are automatically fixed by the latest base lambda image
- `pycryptodome` and `jinja2` were not part of the ticket but are reported by AWS and required only minor version upgrades
- getting `urllib3` upgraded to `2.0.7` was tricky:
  - latest released version for `tableauserverclient` has `urllib3` pinned to `2.0.6`, which means that is incompatible with `2.0.7`, master branch for tableauserverclient is updated to `2.0.7`
  - updated `requirements.in` to use `tableauserverclient` from the master branch but only for `python_version >= 3.11` as we're still using Python 3.8 in the `data-collector`
  - We cannot upgrade `urllib3` to `2.0.7` in `data-collector` without upgrading it to Python `3.11` (the snowflake connector requires urllib3<2 for Python <= 3.10).


Dockerfile
```diff
@@ -73,15 +73,27 @@ RUN yum update -y
 # to use in runtime the git version installed by lambda-git package
 RUN yum install git -y
 
-COPY requirements.txt ./
+COPY requirements.in ./
```
xbeta, line 76: I think the problem is with using `.in` is that the versions aren't pinned.  So we will get a new version at each time this `Dockerfile` is ran. I do not recommend this approach.

mrostan: thanks @xbeta, trying another approach now

delkopiso (approved):
Nice, looks good!

PR #76, opened 2024-01-17
KIL-3014 Hive proxy client

delkopiso:
#### What's up?
- Creates a new `HiveProxyClient` that extends the `BaseDbProxyClient` for use with hive connection types.
- Uses a custom extension of `hive.Connection` (`HiveProxyConnection`) in order to use a custom extension of `hive.Cursor` (`HiveProxyCursor`) which adds an `async_execute` method to wait for incomplete queries as well as canceling queries running longer than the specified timeout.


mrostan (approved):
nice!

PR #77, opened 2024-01-17
KIL-3013 Add ssl support to MySQL proxy client

delkopiso:
#### What's up?
Updates the `MysqlProxyClient` to handle ssl certificate options similarly to how we handle it for `PrestoProxyClient`. This allows us to use the same mysql proxy client for both MySQL transactional databases and Hive MySQL metastores.


mrostan (approved):
nice!

PR #80, opened 2024-01-18
Fix results serialization for Databricks

mrostan:
Updated `DatabricksSqlWarehouseProxyClient` client to extend `BaseDbProxyClient` and implement the right pre-serialization processing for results, as for example dates were not being serialized in the right way (and failed to be serialized when using Azure durable functions)


delkopiso (approved):
nice!

PR #89, opened 2024-02-19
add support to execute scripts

santiagoaguiar:
Introduce a new endpoint `"/api/v1/agent/execute_script/<connection_type>"` to allow execution of python scripts that can access the `connection_type` raw client, a script context, and arguments passed during the call.

The endpoint allows specifying an entry point module that must expose an `execute_script_handler` function that will be used as the entry point for the script. 

It can also include other modules that can be imported from the main module. This is to allow for easier code reuse on the client side, and remove the need of 'transforming' the script to inline those imports. 

At the moment, we require passing the whole script on each call. We can later look into ways of:
* Caching the evaluated script in memory. This could help when script is executed very often, but otherwise, it might get lost during recycling of the lambda containers. As our scripts won't be executed often at the moment, this might not be necessary just yet.
* Retrieving the scripts from the S3 bucket to avoid passing them now. Not clear this would provide advantages as we will still be passing the source from one place to another.

Example API call:
```
curl \
   -H 'Content-Type: application/json' \
   -X POST http://localhost:8081/api/v1/agent/execute_script/snowflake \
   -d @snowflake-execute-script.json
```

`snowflake-execute-script.json`:
```json
{
  "credentials": {
    "connect_args": {
      "user": "<user>",
      "password": "<password>",
      "account": "<account>",
      "warehouse":"<warehouse>"
    }
  },
  "operation": {
      "trace_id": "12346",
      "entry_module":"main",
      "modules": [{ 
        "name":"main", 
        "source":"def execute_script_handler(client, context, sql_query):\n    context.logger.info('running query')\n    if client is None:\n        raise Exception('is none')\n    with client.cursor() as cursor:\n        cursor.execute(sql_query)\n        return {'rows': cursor.fetchmany(10)}"
     }],
     "kwargs": {
     "sql_query": "SELECT 1;"
    },
    "response_type":"url"
  }
}
```


tests/platform_provider.py
santiagoaguiar, line 1: Renamed this file. It caused PyCharm to fail while running tests due to `platform` shadowing the library module. See: https://youtrack.jetbrains.com/issue/PY-23615

apollo/agent/evaluation_utils.py
```diff
@@ -69,6 +73,54 @@ def execute(
...
+    def execute_script(
+        cls,
+        context: Dict,
+        logging_utils: LoggingUtils,
+        operation_name: str,
+        script: AgentScript,
+        trace_id: str,
+    ) -> Optional[Any]:
+        """
+        Executes a list of commands from an operation, returns the result of the
```
mrostan, line 86: nit: this documentation looks wrong

mrostan (approved):
nice! left a single comment with a nit

PR #93, opened 2024-02-27
update looker sdk version

xuanziranhan:
Update looker SDK version. We'd need to pull looker folders based on ID, the old SDK returns error.


sassmith (approved):

mrostan (approved):
looks good

PR #95, opened 2024-03-01
Upgrade dependencies to address cryptography security vulnerability

97nitt:
Get `cryptography` upgraded to address security vulnerability.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -59,8 +59,9 @@ click==8.1.7
     # via
     #   flask
     #   presto-python-client
-cryptography==41.0.7
+cryptography==42.0.5
     # via
+    #   -r requirements.in
     #   azure-identity
     #   azure-storage-blob
     #   msal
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -164,7 +167,7 @@ msal-extensions==1.1.0
     # via azure-identity
 mypy-extensions==1.0.0
     # via typing-inspect
-numpy==1.26.2
+numpy==1.24.4
```
walles2, line 170: We're ok with downgrading Numpy and Pandas here?

97nitt: This is what the dependency resolver came up with 🤷 

I was a little concerned, more with the possibility of introducing a new vulnerability, but `docker scout` scans of the agent images (ran locally) didn't come up with anything new.

97nitt: @walles2 @mrostan any concerns here with the downgraded `numpy` and `pandas`?

mrostan: I think it's fine, data-collector in master is using:
- numpy==1.23.4
- pandas==1.3.5
so, even when we're downgrading we're still using versions after those being used by the dc

mrostan: sorry, assumed it was already approved

mrostan (approved):

PR #96, opened 2024-03-07
[CI] Trufflehog scan is no longer needed in CI

xbeta:
We have the same coverage with github advanced security.


genehynson (approved):

mcd-craig (approved):
LGTM

PR #97, opened 2024-03-14
Support disabling Redshift driver decoding

mrostan:
This feature was added as a workaround for the query to retrieve Redshift query logs failing due to UTF decoding errors.
It basically disables the string decoding in the driver and allow us to decode it later.


mdediana (approved):
Looks good!

pxc-dev (approved):

PR #99, opened 2024-03-19
Close DB Proxy Client's connection on deletion

sassmith:
- Closes the DB Proxy client's connection on deletion of the object
    - FIX: Properly ends Session in Teradata so sessions are not left in `IDLE` state


mrostan (approved):
good finding @sassmith, the change looks good to me
we'll need to review this for the other agent platforms (gcp and aws) as we're caching clients for those platforms, and the client will be deleted only when we try to use it and we detect it is expired, will send more details in the slack [thread](https://montecarlodata.slack.com/archives/C05NWV6LYUU/p1710818919684919)

PR #103, opened 2024-04-02
AWS agent: timeout earlier when getting current image in health endpoint

mrostan:
- As part of the health endpoint we're getting the current image (we get it from CloudFormation or directly from the Lambda if Terraform is being used)
- This operation can take a long time to execute if connections to CloudFormation or Lambda endpoints are not allowed, as the default timeout and retry policies will take between 4 or 5 minutes to fail.
- This PR reduces the timeout used when getting the current image to `10` seconds and a single retry, meaning it would take around 20 seconds to fail. It also updates the health endpoint to succeed even if we weren't able to get the current image, it just includes a new warning message with the error received and a `null` value in the image.
- *Note*: for `lambda` clients the `connect_timeout` setting is not working as expected, and setting `connect_timeout=10` is actually timing out after `160` seconds, that seems to be the case for other values (it times out after value*16), couldn't find this issue reported but I think it might be related to a data type conversion when calling the native `sock.settimeout` function

### DIFF
--- a/apollo/interfaces/lambda_function/aws_utils.py
+++ b/apollo/interfaces/lambda_function/aws_utils.py
@@ -3,9 +3,11 @@ from botocore.config import Config
 
 def get_retrieve_current_image_boto_config(connection_timeout: int) -> Config:
     """
-    Configured the boto3 client (lambda or cloudformation) used to retrieve the current image
-    with 10 seconds timeout and a single retry, so it would take 20 seconds to fail instead
-    of 5 minutes (with the default settings).
+    Returns a boto3 client configuration with the specified connection timeout and a single
+    retry in standard mode.
+    By default, connect_timeout is 60 seconds and legacy retry mode uses 4 attempts, so it takes
+    around 5 minutes to fail if connectivity is not allowed (for example Lambda function configured
+    with no external network access and no VPC endpoints).
     """
     return Config(
         connect_timeout=connection_timeout,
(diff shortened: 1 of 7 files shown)


apollo/interfaces/lambda_function/aws_utils.py
```diff
@@ -0,0 +1,18 @@
+from botocore.config import Config
+
+
+def get_retrieve_current_image_boto_config(connection_timeout: int) -> Config:
```
97nitt, line 4: nit: While the initial use case for this function is for boto client config to retrieve the current image URI from CloudFormation, it could easily serve other use cases. Maybe consider a more generic name (e.g. `get_boto_config`), and allowing `max_attempts` to be passed in (with a default=4).

apollo/interfaces/lambda_function/direct_updater.py
```diff
@@ -125,7 +128,12 @@ def get_current_image(self) -> Optional[str]:
         """
         Returns the current value for `ImageUri` in the function.
         """
-        client = self._get_lambda_client()
+        # in practice, the connect_timeout specified for lambda clients is multiplied by 16
+        # it's not clear why, it could be related to data type conversion when calling
+        # sock.settimeout, so we're setting 1 here to have a 16 seconds timeout.
```
97nitt, line 133: weird

97nitt (approved):

PR #110, opened 2024-04-25
Upload AWS image to CaaS agent repo

mrostan:
Build updated to:
- Move the building of the AWS image to a new step that first builds the image and then uploads it to Docker Hub first and then to ECR (before we were building it first to upload it to Docker and then building it again to upload it to ECR)
- The AWS image is uploaded now to the new ECR repo in the corresponding CaaS account
- Then it is uploaded to the previous ECR repo in dev/prod account, for compatibility reasons, this step will be removed eventually
- Ubuntu images (from circleci) updated to `current` as the previous image we were using is now deprecated

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,7 +128,7 @@ jobs:
       - docker/push:
           image: montecarlodata/<< parameters.docker_hub_repository >>
           tag: latest-lambda,<< parameters.code_version >>-lambda
-      - aws-ecr/login
+      - aws-ecr/ecr_login
       - aws-ecr/push_image:
           repo: << parameters.aws_ecr_repository >>
           tag: << parameters.code_version >>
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -97,37 +91,73 @@ jobs:
...
+          image: montecarlodata/<< parameters.docker_hub_repository >>
+          tag: latest-lambda,<< parameters.code_version >>-lambda
+      - aws-cli/setup:
+          role_arn: "${AWS_ROLE_ARN}"
+      - aws-ecr/ecr_login
+      - aws-ecr/push_image:
           repo: << parameters.aws_ecr_repository >>
-          source-tag: << parameters.code_version >>
+          tag: latest,<< parameters.code_version >>
+  push-lambda-legacy-repo:
```
pxc-dev, line 138: Just so we don't forget cleaning this up, I created a time-bound ticket:  https://linear.app/montecarlodata/issue/KIL-3628/remove-legacy-agent-repo

pxc-dev (approved):

PR #111, opened 2024-05-02
Ssmith/kil 3646 motherduck metadata collection

sassmith:
Add Motherduck support as a db integration

### DIFF
--- a/apollo/integrations/db/motherduck_proxy_client.py
+++ b/apollo/integrations/db/motherduck_proxy_client.py
@@ -1,4 +1,5 @@
 import os
+import duckdb
 from typing import (
     Any,
     Dict,
--- a/apollo/integrations/db/motherduck_proxy_client.py
+++ b/apollo/integrations/db/motherduck_proxy_client.py
@@ -24,7 +24,7 @@ class MotherDuckProxyClient(BaseDbProxyClient):
             raise ValueError(
                 f"Motherduck agent client requires {_ATTR_CONNECT_ARGS} in credentials"
             )
-        self._connection = duckdb.connect(**credentials[_ATTR_CONNECT_ARGS])  # type: ignore
+        self._connection = duckdb.connect(credentials[_ATTR_CONNECT_ARGS])  # type: ignore
 
     @property
     def wrapped_client(self):
--- a/apollo/integrations/db/motherduck_proxy_client.py
+++ b/apollo/integrations/db/motherduck_proxy_client.py
@@ -1,0 +1,31 @@
+from typing import (
+    Any,
+    Dict,
+    List,
+    Optional,
+)
+
+import duckdb
+
+from apollo.integrations.db.base_db_proxy_client import BaseDbProxyClient
+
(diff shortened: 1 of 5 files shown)


apollo/integrations/db/motherduck_proxy_client.py
```diff
@@ -0,0 +1,41 @@
...
+        # Duckdb isn't smart enough for us to specify we only care about connecting to
+        # the cloud. It looks for the env var 'HOME' to know where to set up local files
+        # in case you use it in-memory. Lambda functions don't have a HOME env var so we
+        # must tell Duckdb where to create the local files. This also means that function
+        # memory likely needs to be increased to >1050mb
+        # https://github.com/duckdb/duckdb/issues/3855
+        path = "/tmp"
+        os.environ["HOME"] = path
+        os.makedirs(path, exist_ok=True)
+        self._connection = duckdb.connect(credentials[_ATTR_CONNECT_ARGS])  # type: ignore
```
sassmith, line 37: context on this [here](https://montecarlodata.slack.com/archives/C03BK8E8UU9/p1714660918470949)

Basically we know that on an aws agent the lambda will not have enough memory to run the duckdb connection and memory will need to be increased. We are ok with this limitation for now and will re-evaluate if we find a lot of Motherduck usage

apollo/integrations/db/motherduck_proxy_client.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, credentials: Optional[Dict], **kwargs: Any):
+        if not credentials or _ATTR_CONNECT_ARGS not in credentials:
+            raise ValueError(
+                f"Motherduck agent client requires {_ATTR_CONNECT_ARGS} in credentials"
+            )
+
+        # Duckdb can be used in-memory or to connect to Motherduck, a serverless duckdb.
+        # Duckdb isn't smart enough for us to specify we only care about connecting to
+        # the cloud. It looks for the env var 'HOME' to know where to set up local files
+        # in case you use it in-memory. Lambda functions don't have a HOME env var so we
```
pxc-dev, line 30: Have we confirmed that this approach works on GCP and Azure too? I ask because you explicitly mentioned lambda

mrostan: I agree, we should test this for Azure and GCP.
Also, I think it would be safer to define `HOME` only if not defined yet, I'm pretty sure it's going to be defined for Azure considering they use a Debian image, and maybe also for GCP

mrostan (commented):
left a comment about the HOME env var, similar to Prateek's comment, the rest looks good to me

apollo/integrations/db/motherduck_proxy_client.py
sassmith: I did test with all 3 agent flavors. All 3 did work but GCP and AWS needed the cloudrun/lambda memory increased in order to work. Azure worked without any changes.

I updated the logic to only create HOME env var if it is not already set

mrostan (approved):
looks good to me!

PR #113, opened 2024-05-13
Improve client management to release memory earlier, added connection logs for Redshift

mrostan:
- Added code to clear the context, a cyclic reference was causing the context to be retained until the GC runs, causing the client to be retained too and released later than expected.
- Added code to explicitly close the client (if cache is not used) after the operation completes
- Added log messages when Redshift connections are opened/closed

### DIFF
--- a/apollo/agent/agent.py
+++ b/apollo/agent/agent.py
@@ -480,11 +480,16 @@ class Agent:
             except Exception:  # noqa
                 return AgentUtils.agent_response_for_last_exception(client=client)
             finally:
-                # discard clients that raised exceptions, clients like Redshift keep failing after an error
                 if (response is None or response.is_error) and not operation.skip_cache:
+                    # discard clients that raised exceptions, clients like Redshift keep failing
+                    # after an error
                     ProxyClientFactory.dispose_proxy_client(
                         connection_type, credentials, operation.skip_cache
                     )
+                elif client and operation.skip_cache:
+                    # make sure non-cached clients are closed
+                    logger.info(f"Closing non-cached client: {connection_type}")
+                    client.close()
 
     def _execute_client_operation(
         self,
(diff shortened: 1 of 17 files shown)


apollo/agent/agent.py
```diff
@@ -569,13 +573,16 @@ def _execute(
         }
         context[CONTEXT_VAR_UTILS] = OperationUtils(context)
```
mrostan, line 574: Here we're adding utils to the context, but utils is also referencing the context, creating a cyclic reference.
This is fine and the GC will eventually release it, but it causes the client to be released later than what we need.

apollo/agent/agent.py
```diff
@@ -569,13 +573,16 @@ def _execute(
...
-        return AgentEvaluationUtils.execute(
+        result = AgentEvaluationUtils.execute(
             context,
             self._logging_utils,
             operation_name,
             commands.commands,
             commands.trace_id,
         )
+        # clear cyclic reference involving the client, which delays the release of memory
+        context.clear()
```
mrostan, line 584: this call to clear, clears the dictionary, removing the cycle and releasing the reference to the client

apollo/integrations/db/base_db_proxy_client.py
```diff
@@ -23,8 +23,13 @@ def __init__(self):
 
     # On delete make sure we close the connection
     def __del__(self) -> None:
+        self.close()
+
+    def close(self):
         if self._connection:
+            logger.info("Closing DB Proxy connection")
```
pxc-dev, line 30: Would be useful if we can also log the client type (e.g., redshift etc.)

pxc-dev (approved):

apollo/integrations/db/base_db_proxy_client.py
mrostan: good point, done

PR #120, opened 2024-06-10
Azure Agent improvements: logging, restart and time outs

mrostan:
Multiple improvements to Azure Agents:
- Add more logging to link MC trace ids with Azure Durable Functions operations so we can identify queries causing issues in durable functions
- Fix in `AzureLogContext` to use thread locals to store the context, this could be causing the wrong context to be logged when multiple operations were running at the same time.
- Ability to restart an agent: internally it uses the "update" support and sets a new environment var that is used to force the restart and also to track last time the agent was updated or restarted: `MCD_LAST_UPDATE_TS`
- "Abandon": Durable Functions activities are abandoned after `885` seconds by default, this is intentionally before the Azure Function timeout (configured with `900` seconds) to prevent the task to be automatically retried by Azure and after `860` which is the time out used for MC queries. Based on official Azure docs [here](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-error-handling?tabs=python#function-timeouts).
- Added support to set a new env var that configures Durable Functions: `maxConcurrentOrchestratorFunctions`, it's not clear if we're going to change it but we wanted to have a way to do it in case we need it.

### DIFF
--- a/apollo/interfaces/generic/main.py
+++ b/apollo/interfaces/generic/main.py
@@ -325,16 +325,6 @@ def execute_agent_script(connection_type: str, json_request: Dict) -> AgentRespo
     return agent.execute_script(connection_type, script, credentials)
 
 
-@app.route("/api/v1/test/sleep", methods=["GET"])
-def test_sleep() -> Tuple[Dict, int]:
-    request_dict: Dict = request.args  # type: ignore
-    seconds = int(request_dict.get("seconds", 10))
-    logger.info(f"Waiting for {seconds} seconds")
-    sleep(seconds)
-    logger.info(f"Wait for {seconds} seconds finished")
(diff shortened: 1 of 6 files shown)


apollo/interfaces/generic/main.py
```diff
@@ -2,6 +2,7 @@
 import logging
 import os
 from datetime import datetime, timezone, timedelta
+from time import sleep
```
97nitt, line 5: ☝️ unused?

97nitt (approved):
lgtm, one minor comment re: what I think is probably an unused import.

apollo/interfaces/generic/main.py
mrostan: good catch @97nitt , yes, added that for testing, thanks

PR #124, opened 2024-06-24
Add new endpoint to test DNS lookups

mrostan:
- `/api/v1/test/network/dns?host=HOSTNAME` can now be used to perform a DNS lookup operation and get the IP addresses mapped to the specified hostname

### DIFF
--- a/apollo/agent/agent.py
+++ b/apollo/agent/agent.py
@@ -209,6 +209,34 @@ class Agent:
                 host, port_str, timeout_str, trace_id
             )
 
+    def perform_dns_lookup(
+        self,
+        host: Optional[str],
+        port_str: Optional[str],
+        trace_id: Optional[str] = None,
+    ) -> AgentResponse:
+        """
+        Performs a DNS lookup for the given host name.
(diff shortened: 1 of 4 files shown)


apollo/agent/agent.py
```diff
@@ -209,6 +209,34 @@ def validate_telnet_connection(
...
+        self,
+        host: Optional[str],
+        port_str: Optional[str],
+        trace_id: Optional[str] = None,
+    ) -> AgentResponse:
+        """
+        Performs a DNS lookup for the given host name.
+        :param host: Host to check, will raise `BadRequestError` if None.
+        :param port_str: Optional port to pass to `getaddrinfo` API, both int and
+            string are supported.
```
97nitt, line 222: would it be more accurate for port to be `Union[int, str]`?

97nitt (approved):
Looks good, one minor comment.

apollo/agent/agent.py
mrostan: good catch, updating it

PR #130, opened 2024-07-03
Re-applied vulnerability fixes from 0.5.15 restricting `numpy<2.0.0`

mrostan:
- Applied library upgrades from https://github.com/monte-carlo-data/apollo-agent/pull/126 adding a restriction to `numpy` to be `<2.0.0` and prevent the `numpy.dtype` error described [here](https://github.com/numpy/numpy/issues/26710) that affected `Databricks` connectivity.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,19 +4,19 @@
 #
 #    pip-compile requirements.in
 #
-alembic==1.13.0
+alembic==1.13.2
     # via databricks-sql-connector
 asn1crypto==1.5.1
     # via
     #   oscrypto
     #   snowflake-connector-python
-attrs==23.1.0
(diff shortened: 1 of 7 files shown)


requirements.txt
```diff
@@ -167,38 +167,43 @@ msal-extensions==1.1.0
     # via azure-identity
 mypy-extensions==1.0.0
     # via typing-inspect
-numpy==1.24.4
+numpy==1.26.4
```
mrostan, line 170: `numpy` updated now to `1.26.4` instead of `2.x`

97nitt (approved):

pxc-dev (approved):

PR #131, opened 2024-07-04
ggrushka/DAR-385/tableau url limitation

grgilad:
Tableau API requests are limited to the following format -
 ` url = f"{self._server.baseurl}/sites/{self._server.site_id}/{path}"`

I'm introducing a new URL call which does not have _sites_ in it's path. Since changing the code in it's entirety is a no go (both data-collector and apollo-agent), this allows to send the full path from the dc code.

### DIFF
--- a/apollo/integrations/tableau/tableau_proxy_client.py
+++ b/apollo/integrations/tableau/tableau_proxy_client.py
@@ -138,7 +138,10 @@ class TableauProxyClient(BaseProxyClient):
             "X-Tableau-Auth": self._server.auth_token,
             "Content-Type": content_type,
         }
-        url = f"{self._server.baseurl}/sites/{self._server.site_id}/{path}"
+        if path.startswith("http"):
+            url = path
+        else:
+            url = f"{self._server.baseurl}/sites/{self._server.site_id}/{path}"
         response = requests.request(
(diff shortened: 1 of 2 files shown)


grgilad:
@mrostan can you take a look, see if this hack makes sense?

mrostan (commented):
@grgilad I think this is fine, but just wondering, what if we check for the path starting with "/" and use `{baseurl}{path}` in that case? that way we wouldn't be sending a whole url and potentially connecting to any other site
From the DC code it seems all paths we send are not starting with "/"

mrostan (approved):
lgtm!

grgilad:
> lgtm!

Thanks!

mrostan:
@grgilad just in case you missed it, this is the release process for the agent: https://github.com/monte-carlo-data/apollo-agent?tab=readme-ov-file#release-process
You can use the previous release notes as a reference, we basically use the "generate release notes option" and make a few minor changes

grgilad:
> @grgilad just in case you missed it, this is the release process for the agent: https://github.com/monte-carlo-data/apollo-agent?tab=readme-ov-file#release-process You can use the previous release notes as a reference, we basically use the "generate release notes option" and make a few minor changes

Thanks, will do, want to wait a bit and see if more changes are required.

PR #134, opened 2024-08-01
MES-306 - update boto

walles2:
We're seeing some very strange issues with our Glue Integration (https://linear.app/montecarlodata/issue/MES-494/glue-getdatabases-returning-empty-responses-despite-having-access, https://linear.app/montecarlodata/issue/MES-306/investigate-schema-changing-for-old-mutual, https://linear.app/montecarlodata/issue/MES-460/old-mutual-glue-metadata-collection-timing-out) and want to bump the boto version. There is some difference in the Glue client between the version we're on now and the most updated version.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -1,7 +1,7 @@
 azure-identity==1.17.1
 azure-mgmt-storage==21.2.1
 azure-storage-blob==12.20.0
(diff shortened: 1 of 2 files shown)


mrostan (approved):
lgtm

PR #135, opened 2024-08-15
Add support for Dremio integration

sassmith:
- Add support for Dremio integration
- use Pyarrow FlightClient to execute and fetch results
- Create a description object from the Pyarrow FlightStreamReader schema to return to the collector

### DIFF
--- a/apollo/integrations/db/dremio_proxy_client.py
+++ b/apollo/integrations/db/dremio_proxy_client.py
@@ -1,4 +1,4 @@
-from typing import Optional, Dict, Any, List, Union
+from typing import Optional, Dict, Any, List, Union, Tuple
 
 from pyarrow.flight import FlightDescriptor, FlightCallOptions
 
@@ -22,11 +22,11 @@ class DremioProxyClient(BaseDbProxyClient):
             raise ValueError(
                 f"Dremio agent client requires {_ATTR_CONNECT_ARGS} in credentials"
             )
-        self._connection = flight.connect(location=credentials[_ATTR_CONNECT_ARGS]["host"])  # type: ignore
+        self._connection = flight.connect(**credentials[_ATTR_CONNECT_ARGS])  # type: ignore
         self._headers = [
             (
                 b"authorization",
-                f"bearer {credentials[_ATTR_CONNECT_ARGS]['token']}".encode("utf-8"),
+                f"bearer {credentials.get('token')}".encode("utf-8"),
             )
         ]
 
@@ -58,11 +58,11 @@ class DremioProxyClient(BaseDbProxyClient):
 
         return {
             "records": records,
-            "description": self._set_description(reader.schema),
+            "description": self._get_dbapi_description(reader.schema),
             "rowcount": len(records),
         }
 
-    def _set_description(self, schema: Schema) -> List[List[Any]]:
+    def _get_dbapi_description(self, schema: Schema) -> List[Tuple]:
         """
         Create a DB API compliant description object from the FlightStreamReader schema
         """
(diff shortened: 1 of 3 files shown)


apollo/integrations/db/dremio_proxy_client.py
```diff
@@ -0,0 +1,91 @@
...
+class DremioProxyClient(BaseDbProxyClient):
+    """
+    Proxy client for Dremio. Credentials are expected to be supplied under "connect_args" and
+    will be passed directly to `pyarrow.flight.connect`, so only attributes supported as parameters by
+    `flight.connect` should be passed.
+    """
+
+    def __init__(self, credentials: Optional[Dict], **kwargs: Any):
+        super().__init__(connection_type="dremio")
+        if not credentials or _ATTR_CONNECT_ARGS not in credentials:
```
mrostan, line 21: `connect_args` is intended to be used when we're passing everything in connect_args to a connect method like `connect(**connect_args)`, and the idea is that we can pass new parameters without having to update the agent code
in this case it seems you're getting only `host` and `token` from connect_args, so no need to use it, you can check the code for `Looker` where we're getting `client_id` and `client_secret` directly from credentials
Also, in this case, if `flight.connect` accepts more parameters than host, even when we're not using them now, might be good to keep connect_args and use `flight.connect(**connect_args)` and specify `token` in a separate property in credentials.

makes sense?

apollo/integrations/db/dremio_proxy_client.py
```diff
@@ -0,0 +1,91 @@
...
+            for row in zip(*[column.to_pylist() for column in results.itercolumns()])
+        ]
+
+        return {
+            "records": records,
+            "description": self._set_description(reader.schema),
+            "rowcount": len(records),
+        }
+
+    def _set_description(self, schema: Schema) -> List[List[Any]]:
```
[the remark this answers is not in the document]
mrostan: Also, I think you can convert this method to return `List[Tuple]`, see below

apollo/integrations/db/dremio_proxy_client.py
```diff
@@ -0,0 +1,91 @@
...
+                return 1
+            elif pyarrow_types.is_int32(arrow_type):
+                return 2
+            elif pyarrow_types.is_float32(arrow_type):
+                return 3
+            # Add more type mappings as needed
+            return "unknown"
+
+        return [
+            [
```
mrostan, line 81: I think you can return a tuple here using `()` instead of `[]` which is more dbapi compliant and will be converted automatically to a tuple for serialization anyway

mrostan (commented):
it looks great Sam, left a few minor comments, let me know if you have any questions/comments

apollo/integrations/db/dremio_proxy_client.py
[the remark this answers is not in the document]
sassmith: good on the rename and List[Tuple] 👍

apollo/integrations/db/dremio_proxy_client.py
sassmith: makes total sense! implemented those changes

apollo/integrations/db/dremio_proxy_client.py
```diff
@@ -0,0 +1,91 @@
...
+    will be passed directly to `pyarrow.flight.connect`, so only attributes supported as parameters by
+    `flight.connect` should be passed.
+    """
+
+    def __init__(self, credentials: Optional[Dict], **kwargs: Any):
+        super().__init__(connection_type="dremio")
+        if not credentials or _ATTR_CONNECT_ARGS not in credentials:
+            raise ValueError(
+                f"Dremio agent client requires {_ATTR_CONNECT_ARGS} in credentials"
+            )
```
mrostan, line 24: nit: if `token` is required maybe we should add a validation here too?

mrostan (approved):
looks good, left a minor comment

PR #137, opened 2024-09-06
Ncastagnet/vuln 414 upgrade package go critical

ncastagnet:
- Update the `teradatasql` package to upgrade the included Go version to 1.22.5.


sassmith (approved):
Looks good! Adding here that we tested Teradata connection against this new version in Dev

PR #139, opened 2024-09-27
Update libgnutls30

lexicalunit:
There's two commits here. The first one is just formatting and nothing else. Just for readability. Should hopefully be very straightforward! 🤞

The other commit updates `libgnutls30` which updates the affected library to an unaffected version.

- [x] pushed to `dev` branch


mrostan (approved):
looks good to me, thanks!

PR #140, opened 2024-10-03
VULN-419 upgrade package python from 3.11 to 3.12

romanoff:
Updated python version and setuptools version to eliminate vulnurabilities

Context:
https://linear.app/montecarlodata/issue/VULN-419/upgrade-package-python-high
https://linear.app/montecarlodata/issue/VULN-423/major-upgrade-for-setuptools-high


lexicalunit:
When updating the python version we probably also have to update the `psycopg2-binary` version too.

tests/test_azure_blob_storage_proxy.py
```diff
@@ -304,7 +304,7 @@ def test_download(self, mock_temp_file_path, mock_client_type):
             container=_TEST_BUCKET_NAME, blob=file_key
         )
         self._mock_blob_client.download_blob.assert_called_once_with()
-        self.assertTrue(mock_temp_file_path.called_once())
+        mock_temp_file_path.assert_called_once()
```
lexicalunit, line 307: Nice! It was python 3.12 that flagged these I bet 🤘

lexicalunit (approved):
I recently added some documentation on how to bake apollo changes on `dev` before merging them into production: https://www.notion.so/montecarlodata/Agents-working-on-vulnerability-issues-e240ca5352b04754a1cfbc1632294b28?pvs=4#8a288a33bb3e416ab76913bfb35b8142

Lemme know if it helps or if you'd add anything else to it!

Dockerfile
```diff
@@ -128,22 +128,13 @@ RUN apt update
...
-    && curl https://packages.microsoft.com/config/debian/10/prod.list \
-    > /etc/apt/sources.list.d/mssql-release.list \
-    && apt-get update \
-    && ACCEPT_EULA=Y apt-get install -y --allow-downgrades \
-    msodbcsql17 \
-    odbcinst=2.3.11-1 \
-    odbcinst1debian2=2.3.11-1 \
-    unixodbc-dev=2.3.11-1 \
-    unixodbc=2.3.11-1
+    && apt-get install -y msodbcsql17 odbcinst odbcinst1debian2 unixodbc-dev unixodbc
```
romanoff, line 137: I checked and it installs `2.3.11-3` of odbcinst, odbcinst1debian2, unixodbc-dev, unixodbc in this case. Using `mcr.microsoft.com/azure-functions/python:4-python3.12` `2.3.11-1` option was not available. Should that be fine? @mrostan

lexicalunit: There's a nifty Dockerfile linter that checks for best practice adherence I use called `hadolint`. One thing I recommends is that we pin version for `apt-get install` commands: https://github.com/hadolint/hadolint/wiki/DL3008

This is also recommended by Docker in their [Best practices documentation](https://docs.docker.com/build/building/best-practices/):

> Version pinning forces the build to retrieve a particular version regardless of what’s in the cache. This technique can also reduce failures due to unanticipated changes in required packages.

I think `2.3.11-3` is ok, but I would keep pinning, and just update the version to match what gets installed here.

lexicalunit: Also I think we should keep the `ACCEPT_EULA=Y` part as it's recommended by Microsoft: https://learn.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server?view=sql-server-ver16&tabs=alpine18-install%2Cdebian17-install%2Cdebian8-install%2Credhat7-13-install%2Crhel7-offline

lexicalunit (changes requested):
Nit: We should update version pinnings and keep them.

Issue: We should keep the `ACCEPT_EULA=Y` in accordance with MS installation guide.

lexicalunit (approved):

mrostan (approved):
lgtm

PR #147, opened 2024-10-31
Add HTTP connection validator

mrostan:
- Added a new networking test to validate `HTTP` connections, similar to the `TCP` related operations
- It takes the URL to test, a flag `include_response` indicating if the response should be sent back and the `timeout` in seconds

### DIFF
--- a/apollo/agent/agent.py
+++ b/apollo/agent/agent.py
@@ -253,7 +253,7 @@ class Agent:
         :param trace_id: Optional trace ID received from the client that will be included in
             the response, if present.
         """
-        with self._inject_log_context("perform_dns_lookup", trace_id):
+        with self._inject_log_context("validate_http_connection", trace_id):
             logger.info(
                 "HTTP connection test request received",
                 extra=self._logging_utils.build_extra(
--- a/apollo/agent/agent.py
+++ b/apollo/agent/agent.py
@@ -237,6 +237,41 @@ class Agent:
             )
             return ValidateNetwork.perform_dns_lookup(host, port_str, trace_id)
 
+    def validate_http_connection(
+        self,
+        url: Optional[str],
+        include_response_str: Optional[Union[bool, str]],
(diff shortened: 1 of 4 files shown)


apollo/agent/agent.py
```diff
@@ -237,6 +237,41 @@ def perform_dns_lookup(
...
+        """
+        Performs a GET request to test the provider URL.
+        :param url: The URL to test, will raise `BadRequestError` if None.
+        :param include_response_str: Optional boolean indicating if the response should be sent back.
+        :param timeout_str: Timeout in seconds as a string containing the numeric value,
+            will raise `BadRequestError` if non-numeric. Defaults to 10 seconds.
+        :param trace_id: Optional trace ID received from the client that will be included in
+            the response, if present.
+        """
+        with self._inject_log_context("perform_dns_lookup", trace_id):
```
pxc-dev, line 256: Shouldn't the log context be `test_network_http`?

apollo/validators/validate_network.py
```diff
@@ -192,6 +220,38 @@ def _internal_perform_dns_lookup(
...
+        include_response = (
+            include_response_str == "true" or include_response_str is True
+        )
+        timeout_in_seconds = (
+            int(timeout_str) if timeout_str else _DEFAULT_HTTP_TIMEOUT_SECS
+        )
+
+        try:
+            response = requests.get(url, timeout=timeout_in_seconds)
+            message = f"URL {url} responded with status code: {response.status_code}"
```
pxc-dev, line 245: Same comment as [here](https://github.com/monte-carlo-data/data-collector/pull/1446#discussion_r1825264933).

pxc-dev (approved):

apollo/agent/agent.py
mrostan: yes! thank you!

PR #148, opened 2024-11-01
Support for Hive/Spark on Python 3.12

sassmith:
# What's New
- For Hive and Spark connections the PyHive client does not support python 3.12 when an HTTPS connection is used.
- This PR switches to using the Impyla client which has a beta release that supports python 3.12
- Impyla client's execute() method is already an asynchronous executor.
- A connection timeout parameter is set when the connection is created and is provided to the agent in the connection_args.


mrostan (approved):
lgtm!

PR #150, opened 2024-11-12
VULN-464: Remove sqllite package

benitezfede:
Removed sqllite package from the docker image to fix the VULN issue reported by Aikido


lexicalunit:
Oh wow. Yeah there's no reason to have sqlite 👍🏻 I had no idea we even had that in there.

lexicalunit (approved):

PR #156, opened 2024-12-13
MES-1053 - update databricks sdk

walles2:
The old version of the DBX sdk did not support non-AWS Databricks environments using OAuth. We didn't catch it because we only have AWS Databricks environments for MC to test against (and use in prod).


mdediana (approved):
LGTM

PR #161, opened 2025-01-14
Ssmith/mes 1109 support ssl mysql with the agent

sassmith:
Updates MySQL client to support SSL cert data and maintain support for existing SSL cert path param.

Testing:
Running validations on Azure agent and AWS agent for MySQL with SSL and confirmed the connection was established with SSL fro m the logs:

Screenshot 2025-01-14 at 12 08 09 AM
Screenshot 2025-01-14 at 12 08 09 AM

### DIFF
--- a/apollo/integrations/db/base_db_proxy_client.py
+++ b/apollo/integrations/db/base_db_proxy_client.py
@@ -52,6 +52,7 @@ class SslOptions:
     verify_cert: bool = True
     verify_identity: bool = True
     disabled: bool = False
+    mechanism: str | None = None
 
     def __post_init__(self):
         # Validation for conflicting flags
--- a/apollo/integrations/db/base_db_proxy_client.py
+++ b/apollo/integrations/db/base_db_proxy_client.py
@@ -106,6 +106,9 @@ class SslOptions:
 
     def _set_cert_and_key_to_context(self, ssl_context: ssl.SSLContext) -> None:
         """Check if temp file exists, if not create it from certificate or key data."""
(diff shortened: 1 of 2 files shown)


benitezfede (approved):
LGTM. We can think if it is worth to move the `SslOptions` class to common so we do not have it duplicated.

apollo/integrations/db/base_db_proxy_client.py
```diff
@@ -17,6 +20,111 @@
 logger = logging.getLogger(__name__)
 
 
+@dataclass
```
mdediana, line 23: Nit: I understand we can't reuse collector code in the agent, but can we reuse agent code in the collector? If so we could delete this class in the other codebase.

mdediana (approved):
LGTM

apollo/integrations/db/base_db_proxy_client.py
sassmith: @mrostan do you know of an example in the data-collector where we are doing this already? I am not sure how that would work

mrostan: sure, you can check the implementation for `AzureBlobReaderWriter` [here](https://github.com/monte-carlo-data/data-collector/blob/00b518c890c2ca5a89466cdd0564b8cdbfa09949/lambdas/clients/azure_blob/reader_writer.py#L14), it extends a class defined in `apollo-agent`.

Remember we have a dependency to the agent, we update it mainly to update libraries but we can also use code there.

PR #169, opened 2025-02-14
Add Optional openSSL Config to Agent Resources

sassmith:
Creates a `resources/lambda/openssl` directory for optional openSSL config files


mrostan (approved):
lgtm

PR #170, opened 2025-02-19
add support for an s3 proxy client

santiagoaguiar:
At the moment we need to download `athena` results by making calls to get_query_results. The maximum page size to return is 1000, which is too low when fetching million of results (eg. for metric monitors with very high segment count), requiring too many roundtrips and long execution times to execute a monitor run or backfill job.

Add an S3 proxy client to support downloading athena result files from S3 and reduce the number of roundtrips. 

I'm adding a separate proxy client instead of using the `StorageClient`, as storage client API is geared to represent the MC storage agent specifically.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ repos:
     hooks:
     - id: black
       name: black
-      entry: ./.venv/bin/black
+      entry: black
       language: python
       types_or: [python, pyi]
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ repos:
     hooks:
     - id: black
       name: black
-      entry: black
+      entry: ./.venv/bin/black
       language: python
       types_or: [python, pyi]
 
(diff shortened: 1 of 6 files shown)


apollo/integrations/aws/s3_proxy_client.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class S3ProxyClient(BaseAwsProxyClient):
+    def __init__(self, credentials: Optional[Dict], **kwargs: Any):
+        BaseAwsProxyClient.__init__(self, service_type="s3", credentials=credentials)
+
+    def process_result(self, value: Any) -> Any:
+        """
+        Process the result of the methods on this client before being serialized to JSON.
+        Serializes the value to JSON using the AgentSerializer and parses it back from JSON.
```
mrostan, line 15: this comment is no longer accurate, right?

.pre-commit-config.yaml
```diff
@@ -3,7 +3,7 @@ repos:
     hooks:
     - id: black
       name: black
-      entry: ./.venv/bin/black
+      entry: black
```
mrostan, line 6: this was here to support Git UI clients that doesn't have the environment you have in the CLI, in the README we ask to have the venv at that location or create a symlink
But I understand this is not standard among other projects.

mrostan (approved):

.pre-commit-config.yaml
santiagoaguiar: OK, I will revert.

apollo/integrations/aws/s3_proxy_client.py
santiagoaguiar: Right, updated!

PR #172, opened 2025-02-28
Upgrade cryptography lib

jeremyhewett:


mdediana (approved):
LGTM

There is a difference regarding strip-extras, I'd try not to change it, but really not sure about the difference.

jeremyhewett:
When I ran it without strip-extras I got a warning saying it's going to become the default and they recommend including it, so I added it. I think it's fine, I see it in our other repos.

PR #174, opened 2025-03-11
add Teradata SSL support using SslOptions, add unit tests

sassmith:
- Add support for Teradata SSL connections
- Add method to SslOptions that writes ca-data to a temp file to be passed to the Teradata client
- Adds unit test


mdediana (approved):
LGTM

PR #175, opened 2025-03-11
feat: aws_generic platform

genehynson:
- adds support for a new platform, AWS_GENERIC, which allows customers to run the agent on the AWS infra of their choice (e.g. EKS, ECS, EC2)
- the agent can be run via the Docker image


.pre-commit-config.yaml
```diff
@@ -3,7 +3,7 @@ repos:
     hooks:
     - id: black
       name: black
-      entry: ./.venv/bin/black
+      entry: black
```
genehynson, line 6: was there a reason for this? removing the .venv path makes this work easier with pyenv

apollo/interfaces/aws_metal/platform.py
```diff
@@ -0,0 +1,82 @@
...
+        limit: int,
+    ) -> Dict | None:
+        try:
+            return super().filter_log_events(
+                pattern, start_time_str, end_time_str, limit
+            )
+        except AgentConfigurationError:
+            # if env var is not set, log collection is disabled for AWS Metal
+            logger.info("Log collection is disabled for this agent")
+            return None
```
genehynson, line 64: silence errors about log collection not configured for AWS_METAL platform

apollo/interfaces/generic/platforms/aws_generic/platform.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @property
+    def updater(self) -> None:
+        # automatic updates are not supported for AWS generic deployments
+        return None
+
+    def get_infra_details(self) -> Dict:
+        return {}
+
+    def filter_log_events(
```
genehynson, line 53: I think I need to do a bit more refactoring to expose these functions to the agent's API

genehynson: I think you're right - I'm about to commit a refactoring to do that

apollo/interfaces/generic/main.py
```diff
@@ -1169,6 +1173,27 @@ def get_outbound_ip_address() -> Tuple[Dict, int]:
     return response.result, response.status_code
 
 
+@app.route("/")
```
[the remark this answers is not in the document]
genehynson: I don't have a strong preferences - some examples I found online used the root route as the default health check endpoint so I did that here too. but we could easily use `/health` instead

apollo/integrations/storage/base_storage_client.py
```diff
@@ -205,3 +207,6 @@ def _remove_prefix_from_entries(
             {**entry, "Key": self._remove_prefix(cast(str, entry.get("Key")))}
             for entry in entries
         ]
+
+    def _is_presigned_url_enabled(self) -> bool:
+        return os.getenv(VALIDATE_PRE_SIGNED_URLS_ENV_VAR, "true").lower() == "true"
```
genehynson, line 212: see https://github.com/monte-carlo-data/data-collector/pull/1679#discussion_r1999641591

apollo/interfaces/generic/main.py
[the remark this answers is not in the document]
genehynson: I added both

apollo/interfaces/generic/platforms/aws_generic/platform.py
genehynson: refactored. now lambda + aws_generic both use the same `main.py`

.circleci/config.yml
```diff
@@ -100,6 +100,12 @@ jobs:
           extra_build_args: --target generic --build-arg code_version=<< parameters.code_version >> --build-arg build_number=<< pipeline.number >>
           image: montecarlodata/<< parameters.docker_hub_repository >>
           tag: latest-generic,<< parameters.code_version >>-generic
+      - docker/build:
```
mrostan, line 103: Just wondering, should we upload this image to a new ECR repo too?
Can you use an image directly from DockerHub on ECS/EKS? Asking because I have always used ECR

apollo/interfaces/aws_generic/main.py
mrostan, line 1: it seems we're using this file: `aws_generic/main.py` for both AWS Generic and Lambda, should it be just `aws/main.py` ?

apollo/interfaces/lambda_function/handler.py
```diff
@@ -3,34 +3,17 @@
...
 logging.setLoggerClass(ExtraLogger)
 
 from apig_wsgi import make_lambda_handler
 
-from apollo.agent.env_vars import (
-    DEBUG_ENV_VAR,
-)
-from apollo.interfaces.lambda_function.main import main
+from apollo.interfaces.aws_generic.main import main
 from apollo.interfaces.lambda_function.platform import AwsPlatformProvider
```
mrostan, line 14: let's confirm in dev we still get structured logs properly in the lambda agents with this change

mrostan (approved):
it looks good to me, left a few questions/suggestions

.circleci/config.yml
mrostan: nice, in that case I think there's no need to upload it to ECR

apollo/interfaces/lambda_function/handler.py
genehynson: AWS remote agent logs after upgrading to latest dev version:

image

genehynson: AWS generic remote agent logs after upgrading to latest dev version:
image

PR #178, opened 2025-03-19
chore: change default GUNICORN_WORKERS

genehynson:
removes duplicate lines and adjust default GUNICORN_WORKERS to 4


Dockerfile
```diff
@@ -1,12 +1,7 @@
 FROM python:3.12.9-slim AS base
 
 # Web server env var configuration
-ENV GUNICORN_WORKERS=1
-ENV GUNICORN_THREADS=8
-ENV GUNICORN_TIMEOUT=0
-
-# Web server env var configuration
-ENV GUNICORN_WORKERS=1
+ENV GUNICORN_WORKERS=4
```
mrostan, line 4: should we use 5 so we can handle 40 concurrent requests? closer to the 42 supported by lambda? or do you think that would be too much for a single instance

mrostan (approved):
looks good, left just a thought

Dockerfile
genehynson: oh right I forgot about the magic 42 number! will change it to 5. that actually more closely aligns with Gunicorn's recommendation of `2 * CPU + 1` https://docs.gunicorn.org/en/latest/design.html#how-many-workers

PR #180, opened 2025-03-27
Set Tableau verify_ssl parameter when making direct API calls

sassmith:
- Allow setting Tableau `verify_ssl` connection parameter when making direct API calls


mrostan (approved):
lgtm

PR #183, opened 2025-04-10
feat: merge connect_args from dc and external creds

genehynson:
in the DC we specify some hard coded values that we want to include in `connect_args`. instead of requiring customers to include these hard coded values in their self-hosted creds, lets merge the `connect_args` objects from the DC and provided by the customer

context: https://montecarlodata.slack.com/archives/C083XJFGN2X/p1744323378271009

- [x] Unit tests
- [x] Deployed to dev

### DIFF
--- a/apollo/credentials/asm.py
+++ b/apollo/credentials/asm.py
@@ -22,7 +22,11 @@ class AwsSecretsManagerCredentialsService(BaseCredentialsService):
                 raise ValueError(
                     f"Failed to fetch credentials from AWS Secrets Manager: No secret string found for secret name: {secret_name}"
                 )
-            return json.loads(secret_str)
+            external_credentials = json.loads(secret_str)
+            return self._merge_connect_args(
+                incoming_credentials=credentials,
+                external_credentials=external_credentials,
+            )
         except Exception as e:
             raise ValueError(
                 f"Failed to fetch credentials from AWS Secrets Manager: {e}"
--- a/apollo/credentials/asm.py
+++ b/apollo/credentials/asm.py
@@ -11,7 +11,7 @@ class AwsSecretsManagerCredentialsService(BaseCredentialsService):
     Credentials service that fetches credentials from AWS Secrets Manager.
     """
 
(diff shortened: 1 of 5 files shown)


apollo/credentials/base.py
```diff
@@ -6,3 +6,34 @@ class BaseCredentialsService:
...
+        if not external_connect_args:
+            # when external credentials don't have connect args, return incoming connect args
+            external_credentials["connect_args"] = incoming_connect_args
+            return external_credentials
+        if not isinstance(external_connect_args, dict) or not isinstance(
+            incoming_connect_args, dict
+        ):
+            # when connect args is not a dict (could be a connection string), return
+            # external credentials as is
+            return external_credentials
```
genehynson, line 32: some clients send `connect_args` as a string instead of a dict (e.g. SQL Server), in this case we can't merge them so just use what the customer has provided

apollo/credentials/asm.py
```diff
@@ -22,7 +22,11 @@ def get_credentials(self, credentials: dict) -> dict:
                 raise ValueError(
                     f"Failed to fetch credentials from AWS Secrets Manager: No secret string found for secret name: {secret_name}"
                 )
-            return json.loads(secret_str)
+            external_credentials = json.loads(secret_str)
+            return self._merge_connect_args(
```
mrostan, line 26: just a comment, take it or leave it, it's just a matter of preferences

in a case like this you can define `get_credentials` in the base class (`BaseCredentialService`) as:
```
external_credentials = self._load_external_credentials()
return self._merge_connect_args(
   incoming_credentials=credentials,
   external_credentials=external_credentials,
)
```
and `load_external_credentials` is defined as an abstract method in the base class and implemented in sub-classes

That way you have a little bit less of risk on having a new subclass that forgets to call `merge_connect_args`.
Again, minor

mrostan (approved):
looks good, just a nit, feel free to ignore

apollo/credentials/asm.py
genehynson: good suggestion!

PR #193, opened 2025-06-23
MES-1606: Add Salesforce Data Cloud proxy client

jeremyhewett:
This PR adds a new client: Salesforce Data Cloud. It uses the salesforce-cdp-connector library to extract data from Salesforce Data Cloud.


apollo/integrations/db/salesforce_data_cloud_proxy_client.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+
+    @property
+    def wrapped_client(self):
+        return self._connection
+
+    def close(self):
+        self._connection.close()
+
+    def list_tables(self) -> list[dict]:
```
jeremyhewett, line 41: This custom method is used for metadata collection. The SalesforceCDPConnection library provides a cursor implementation but only for data queries, not metadata.

PR #195, opened 2025-06-26
Fix Salesforce Data Cloud factory function.

jeremyhewett:


apollo/integrations/db/salesforce_data_cloud_proxy_client.py
```diff
@@ -7,13 +7,13 @@
 class SalesforceDataCloudCredentials:
     def __init__(
         self,
-        host: str,
+        domain: str,
```
jeremyhewett, line 10: Salesforce CRM uses `domain` for the salesforce URL. Having them use the same name makes sense since they're the same value.

genehynson (approved):

PR #196, opened 2025-06-26
MES-1505: Salesforce-crm minimum query language

toffer:
https://linear.app/montecarlodata/issue/MES-1505/monolith-salesforce-crm-minimum-query-language

Adds support for performing count queries and row limit queries for Salesforce CRM.

Salesforce Object Query Language (SOQL) doesn't allow us to transform count and row limit queries the typical way, where we run the original query as a subquery that we wrap a `SELECT COUNT(*)` or `LIMIT` clause around. Instead, we run the query as-is, and then modify the result rows to return either the count, or return the subset of rows.

The logic for determining whether to run `execute()` or `execute_count_query()` or `execute_row_limit()` is found on the data-collector side.

### DIFF
--- a/apollo/integrations/db/salesforce_crm_proxy_client.py
+++ b/apollo/integrations/db/salesforce_crm_proxy_client.py
@@ -27,7 +27,7 @@ class SalesforceCRMProxyClient(BaseDbProxyClient):
         description = []
         records = []
 
-        results = self._connection.query(query)
+        results = self._connection.query_all(query)
         rowcount = results["totalSize"]
         records_raw = results["records"]
 
@@ -58,8 +58,13 @@ class SalesforceCRMProxyClient(BaseDbProxyClient):
         description = []
         records = []
 
-        # limit cannot be negative.
-        limit = max(limit, 0)
+        # Return early, if we're not going to return any rows.
+        if limit < 1:
+            return {
+                "description": description,
+                "records": records,
+                "rowcount": 0,
+            }
 
         # REST API docs say batch size must be between 200 and 2000.
         if limit < 200:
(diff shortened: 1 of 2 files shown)


apollo/integrations/db/salesforce_crm_proxy_client.py
```diff
@@ -39,6 +39,43 @@ def execute(self, query: str) -> Dict:
 
         return {"description": description, "records": records, "rowcount": rowcount}
 
+    def execute_count_query(self, query: str) -> Dict:
+        headers = {
+            "Sforce-Query-Options": "batchSize=200",
```
mdediana, line 44: Nit: Why 200? Can we leave a comment, even if the answer is "it's an arbitrary value"?

apollo/integrations/db/salesforce_crm_proxy_client.py
```diff
@@ -39,6 +39,43 @@ def execute(self, query: str) -> Dict:
...
+            "description": description,
+            "records": records,
+            "rowcount": len(records),
+        }
+
+    def execute_row_limit(self, query: str, limit: int) -> Dict:
+        description = []
+        records = []
+
+        limit = min(limit, 200)
```
mdediana, line 60: Any concern with validation? For example, if `limit = 0` or `limit = -1`?

mdediana (approved):
Nice!

apollo/integrations/db/salesforce_crm_proxy_client.py
```diff
@@ -39,6 +39,43 @@ def execute(self, query: str) -> Dict:
 
         return {"description": description, "records": records, "rowcount": rowcount}
 
+    def execute_count_query(self, query: str) -> Dict:
+        headers = {
+            "Sforce-Query-Options": "batchSize=200",
+        }
+        results = self._connection.query(query, headers=headers)
```
toffer, line 46: I'm going to have to rework both `execute()` and `execute_row_limit()` to use `query_all_iter()` instead of `query()` so I can iterate through all results when there are a lot of rows returned. 

(For `execute_count_query()` I get the totalSize in the first batch and throw away the result rows, so I can get away with using `query()` and getting a single batch of results.)

apollo/integrations/db/salesforce_crm_proxy_client.py
toffer: Done in 4788059bcd582edf9e186d4cf00be869548ad0ce.

PR #212, opened 2025-09-15
Vuln-658 - remove sqlite

walles2:
This removes sqlite from our docker image, as nothing in our system uses it, and removing all of the vulns that sqlite comes with.


mdediana (approved):
LGTM

PR #218, opened 2025-09-26
Upgrade gunicorn

mdediana:
Even though a ticket hasn't been created, this is showing in Aikido.

Always upgrading `pip` is an attempt to see less VULNs.


mrostan (approved):
lgtm

PR #267, opened 2026-04-07
feat(ctp): Phase 2 — register all connectors and remove legacy dual-path code

walles2:
## Summary

Phase 2 of the Credential Transform Pipeline (CTP) migration: all remaining connectors are now registered with CTP default configs, and the legacy dual-path credential construction code has been removed from proxy constructors.

- **Register all connectors**: Starburst Galaxy, Starburst Enterprise, SAP HANA, Salesforce CRM, Redshift, Postgres, MySQL, Oracle, Hive, Snowflake, BigQuery, Databricks, Teradata, DB2, Presto, Dremio, Motherduck, Git, HTTP, AWS, Looker, Power BI, Tableau, SQL Server, Fabric
- **Remove legacy dual-path code**: Proxy constructors no longer contain `if connect_args / else raw` branching — they read exclusively from `connect_args`
- **DC pre-shaped credentials**: `CtpRegistry.resolve()` unwraps `{"connect_args": {...}}` and re-runs the pipeline, so both flat and DC-pre-shaped credentials follow the same transform path
- **New Databricks token resolver**: `resolve_databricks_token` transform handles PAT and OAuth credential shapes
- **Comprehensive tests**: 366 tests passing; every connector has pipeline correctness + credential safety + log safety coverage

## Testing

- All 366 unit tests pass
- Local DC→agent integration tested against dev instances: Postgres, Databricks (PAT + OAuth), Snowflake, Redshift, MySQL, SQL Server, Teradata
- Deployed to dev and monitored: no CTP-related errors in Datadog or Sentry across the most-integrated agent (uuid `5c809970-6fa7-4cf3-9d6c-07a47c70fd28`)

## Notes

- Phase 3 (DC-side changes to stop pre-shaping `connect_args`) is tracked separately and not part of this PR
- The `CtpRegistry.resolve()` pass-through for unregistered types is preserved — unrecognised connection types are not affected


walles2:
### Code review

Found 1 issue:

1. `apollo/integrations/ctp/CLAUDE.md` is factually incorrect after this change. The Key Concepts section still says *"If `creds` already contain `connect_args`, the pipeline is skipped (legacy / pre-transformed path)."* But `CtpRegistry.resolve()` now **unwraps** the inner dict and re-runs the pipeline through `CtpPipeline().execute()` — the opposite of skipping. Future contributors relying on this doc to understand the pre-shaped path will be misled.

https://github.com/monte-carlo-data/apollo-agent/blob/fdf6c88f69892347994d01283ab4c62687b6e512/apollo/integrations/ctp/CLAUDE.md#L13-L16

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

walles2:
### Additional findings (scored 75 — slightly below threshold but worth addressing or explicitly dismissing)

**1. Databricks SQL Warehouse — `server_hostname` fallback removed with no CTP equivalent**

The old proxy client derived `server_hostname` from `databricks_workspace_url` when the former was absent:
```python
if SERVER_HOSTNAME not in connect_args:
    connect_args[SERVER_HOSTNAME] = AgentUtils.normalize_url(
        connect_args.get("databricks_workspace_url"), with_scheme=False
    )
```
The new CTP mapper has no fallback — it requires `server_hostname` as `Required[str]`. Any customer credential that historically sent only `databricks_workspace_url` will now get `server_hostname=None` and fail to connect.

https://github.com/monte-carlo-data/apollo-agent/blob/fdf6c88f69892347994d01283ab4c62687b6e512/apollo/integrations/ctp/defaults/databricks.py#L31-L36

**2. Redshift — `autocommit` is now unconditional**

Before this PR, autocommit was opt-in (`if credentials.get("autocommit", False)`). It is now always `True`. Psycopg2 defaults to `autocommit=False`, so this silently changes transaction semantics for any Redshift connection that did not previously set `autocommit: true`. The class docstring still says the old conditional behavior. Either this is intentional (in which case the docstring needs updating and a note in the PR would be helpful) or it's an unintended regression.

https://github.com/monte-carlo-data/apollo-agent/blob/fdf6c88f69892347994d01283ab4c62687b6e512/apollo/integrations/redshift/redshift_proxy_client.py#L14-L25

**3. SQL Server — Phase 2 timeout migration left incomplete**

The `SqlServerOdbcArgs` TypedDict has an explicit note saying Phase 2 would move `login_timeout` and `query_timeout_in_seconds` into `connect_args`. This PR completes Phase 2 for SQL Server but the proxy client still reads these from the top-level credential dict. The comment should either be removed (if the intent is to defer this indefinitely) or the migration completed.

https://github.com/monte-carlo-data/apollo-agent/blob/fdf6c88f69892347994d01283ab4c62687b6e512/apollo/integrations/ctp/defaults/sql_server.py#L16-L20

**4. Databricks REST — `resolve_databricks_token` step will raise on DC pre-shaped credentials**

The `DATABRICKS_REST_DEFAULT_CTP` step has no `when` guard. When the registry unwraps a DC pre-shaped `{"connect_args": {"token": "...", "databricks_workspace_url": "..."}}` and re-runs the pipeline, the transform receives `raw.databricks_token = None`, `raw.client_id = None`, and `raw.client_secret = None`, and raises `CtpPipelineError("No supported Databricks credentials found")`. The DC pre-shaped path for `databricks-rest` is not tested. A `when: raw.databricks_token is defined or raw.client_id is defined` guard (or a pass-through for `raw.token`) would fix this.

https://github.com/monte-carlo-data/apollo-agent/blob/fdf6c88f69892347994d01283ab4c62687b6e512/apollo/integrations/ctp/transforms/resolve_databricks_token.py#L126-L136

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

walles2:
### Code review

Found 1 issue:

1. Dremio `location` field_map fails with `UndefinedError` on DC pre-shaped credentials — `raw.host` and `raw.port` are undefined when the DC sends `{"connect_args": {"location": "grpc+tls://host:port"}}`, since `CtpRegistry.resolve()` unwraps `connect_args` and passes the inner dict as `raw`, leaving only `raw.location` defined. With `StrictUndefined`, this raises at runtime for any Dremio deployment using the DC pre-shaped credential format (the format demonstrated in `tests/test_dremio_client.py`).

https://github.com/monte-carlo-data/apollo-agent/blob/64de48f10ee3ed5e27ee85217d6cf2fdeb5f1cbe/apollo/integrations/ctp/defaults/dremio.py#L24-L29

Fix: add a fallback — e.g. `{{ raw.location if raw.location is defined else (('grpc+tls' if raw.use_tls is defined and raw.use_tls else 'grpc') ~ '://' ~ raw.host ~ ':' ~ raw.port) }}`

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

apollo/integrations/redshift/redshift_proxy_client.py
```diff
@@ -13,15 +13,14 @@ class RedshiftProxyClient(PostgresProxyClient):
...
+    autocommit is always enabled on the connection.
     """
 
     def __init__(self, credentials: Optional[Dict], **kwargs):  # type: ignore
         PostgresProxyClient.__init__(
             self, credentials=credentials, client_type="redshift"
         )
-        if credentials and credentials.get("autocommit", False):
-            self._connection.autocommit = True
+        self._connection.autocommit = True
```
sassmith, line 23: this is a behavior change? previously connections defaulted to transactional mode unless explicitly opted in.

apollo/integrations/tableau/tableau_proxy_client.py
```diff
@@ -112,21 +96,14 @@ def __init__(self, credentials: Optional[Dict], **kwargs: Any):  # noqa
...
-        token = generate_jwt(
-            user_name=self._user_name,
-            client_id=self._client_id,
-            secret_id=self._secret_id,
-            secret_value=self._secret_value,
-            expiration_seconds=expiration_seconds,
-        )
-        auth = JwtAuth(token=token, site_id=self._site_name)
+    def _sign_in(self):
+        auth = JwtAuth(token=self._token, site_id=self._site_name)
```
sassmith: so previously we would create a new 5-minute JWT on every _sign_in() call, called beefed each metadata_query and api_request. This change will be a 5-minute JWT generated once during CTP pipeline execution and reused for all subsequent _sign_in() calls. Will this be an issue for a long running tableau collection job?

apollo/integrations/databricks/databricks_rest_proxy_client.py
```diff
@@ -1,95 +1,26 @@
...
 
     def __init__(self, credentials: Optional[Dict], **kwargs: Any):
-        # Support both flat and connect_args-wrapped formats
-        creds: Dict = {}
-        if credentials:
-            creds = dict(credentials.get(_ATTR_CONNECT_ARGS, credentials))
-
-        token = self._get_token(creds)
-        self._http_client = HttpProxyClient(credentials={"token": token})
+        creds = credentials["connect_args"] if credentials else {}
```
sassmith, line 20: nit: if creds is `{}` should we just fail fast instead of sending a request with `Bearer None`?

walles2:
Good catch. This is intentional — the DC has always hardcoded `autocommit: true` in the Redshift credentials it sends, so the conditional was always evaluating to `True` in practice. Making it unconditional removes the now-vestigial credential field and makes the behavior explicit. The docstring has been updated to reflect this.

apollo/integrations/redshift/redshift_proxy_client.py
walles2: Confirmed not a regression — the DC unconditionally sets `autocommit: True` in both code paths:

- **Agent path** (`lambdas/clients/redshift/wrapper.py:109`): `agent_credentials = {"connect_args": connect_args, "autocommit": True}`
- **Non-agent path** (`lambdas/clients/redshift/wrapper.py:126`): `self._connection.autocommit = True`

The old conditional in the proxy client (`if credentials.get("autocommit", False)`) was always evaluating to `True`. Hardcoding it here is just making that implicit invariant explicit.

sassmith (approved):

PR #272, opened 2026-04-14
Add the ability to pass through and validate custom CTP configs

walles2:
## Summary

- Threads a per-request `ctp_config` through `execute_operation` → `ProxyClientFactory`, enabling custom Jinja2 credential transform pipelines to be passed from the monolith alongside each DC operation
- Adds a `POST /api/v1/ctp/validate/<connection_type>` endpoint for structural validation of custom CTP configs (deserialization, transform type existence, Jinja2 syntax, TypedDict key coverage) without executing against real credentials
- Adds `connect_args_defaults` to `CtpConfig` so connector-level constants (keepalives, timeouts, `http_scheme`, etc.) are injected under the mapper output — ensuring custom CTPs don't silently drop required driver params

## Key Decisions

- **Client cache bypassed when `ctp_config` is present** — two requests with identical credentials but different CTP configs would otherwise share a cached client built from the wrong resolved credentials
- **`connect_args_defaults` merge strategy: mapper wins** — the defaults are applied under the mapper output so a custom config can always override them; registered pipeline defaults are inherited into custom configs via `CtpRegistry.resolve_custom`
- **Unknown keys in `field_map` are errors** — unknown keys passed through silently to the driver before, masking typos like `passwordx`; now rejected at both map-time and validation-time
- **Step-covered missing required keys are warnings, not errors** — if `config.steps` is non-empty, a key missing from the top-level `field_map` may be provided at runtime by a transform step, so `valid` stays `True`
- **Jinja2 syntax validation uses `SandboxedEnvironment`** — prevents template expressions from accessing arbitrary Python internals during the structural check

## Test plan

- [ ] Unit tests cover: `CtpConfig`/`MapperConfig` deserialization, flat mapper shorthand, `connect_args_defaults` merge, unknown-key rejection, validator endpoint happy/error paths, cache bypass
- [ ] E2E: configure a custom CTP config on a connection in dev, verify credentials are transformed and DC operations succeed
- [ ] Validate endpoint: send a malformed config, confirm structured errors are returned

### DIFF
--- a/apollo/agent/agent.py
+++ b/apollo/agent/agent.py
@@ -449,7 +449,7 @@ class Agent:
         operation_name: str,
         operation_dict: Optional[Dict],
         credentials: Optional[Dict],
-        custom_ctp: Optional[Dict] = None,
+        ctp_config: Optional[Dict] = None,
     ) -> AgentResponse:
         """
         Executes an operation for the given connection type using the provided credentials.
@@ -480,7 +480,7 @@ class Agent:
             credentials,
             operation,
             lambda client: self._execute(client, operation_name, operation),
-            custom_ctp=custom_ctp,
+            ctp_config=ctp_config,
         )
 
     def execute_script(
@@ -534,7 +534,7 @@ class Agent:
         credentials: Optional[Dict],
         operation: AgentOperation,
         func: Callable[[BaseProxyClient], Any],
-        custom_ctp: Optional[Dict] = None,
+        ctp_config: Optional[Dict] = None,
     ):
         with self._inject_log_context(
             f"{connection_type}/{operation_name}", operation.trace_id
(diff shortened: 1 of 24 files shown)


walles2:
## Review fixes

Addressed F1 from code review:

**F1 — Cache bypass for custom CTP** (`proxy_client_factory.py`): Added `or ctp_config` to the `skip_cache` branch condition so requests with a custom `ctp_config` always create a fresh client instead of potentially hitting a cached client resolved with different credentials.

walles2:
**F3 — Redshift connect_timeout** (): Raised the default from 5s to 30s. 5 seconds was too tight for connections over slow PrivateLink paths; 30 seconds is a safer default that still bounds connection hangs.

benitezfede (approved):
LGTM

PR #273, opened 2026-04-14
feat(ctp): add resolve_redshift_credentials transform

walles2:
## What this enables

Adds a new CTP transform — `resolve_redshift_credentials` — that exchanges IAM identity for short-lived Redshift database credentials via `redshift:GetClusterCredentials`. This is a building block for Redshift connector CTP pipelines where the agent needs to federate an IAM user into a Redshift DB user rather than use a static password.

**What you can now do in a CTP pipeline:**

```json
{
  "type": "resolve_redshift_credentials",
  "input": {
    "cluster_identifier": "{{ raw.cluster_identifier }}",
    "db_user": "{{ raw.db_user }}",
    "db_name": "{{ raw.db_name }}",
    "aws_region": "{{ raw.aws_region }}",
    "assumable_role": "{{ raw.assumable_role | default(none) }}",
    "duration_seconds": "{{ raw.duration_seconds | default(none) }}"
  },
  "output": {
    "user": "redshift_user",
    "password": "redshift_password"
  }
}
```

**Supported inputs:**
- Required: `cluster_identifier`, `db_user`, `db_name`, `aws_region`
- Optional: `assumable_role` (ARN), `external_id` (for cross-account role assumption), `duration_seconds` (900–3600)

**Outputs** land in `state.derived` under the keys specified in `output.user` and `output.password`, ready to be consumed by the mapper.

## Key Decisions

- **`auto_create` excluded** — omitted intentionally; auto-creating Redshift users from a pipeline step is a footgun (silent side-effect on auth failure), so the transform requires the DB user to already exist.
- **Follows existing boto3 role-assumption pattern** — same STS `assume_role` + session construction used by `base_aws_proxy_client.py` and `resolve_databricks_token`.
- **Inline imports for `time` and `AgentUtils`** — kept inside the `if assumable_role:` branch (matching the established pattern in other AWS transforms) to avoid importing agent internals at module load time on the non-role path.

## Test plan

- [x] 12 unit tests covering: happy path, role assumption with and without external ID, optional `duration_seconds`, custom output keys, all required input/output validation errors, API error wrapping, and registry registration
- [x] All 360 CTP tests passing
- [x] Typecheck clean (pyright 0 errors)

### DIFF
--- a/apollo/integrations/ctp/transforms/resolve_redshift_credentials.py
+++ b/apollo/integrations/ctp/transforms/resolve_redshift_credentials.py
@@ -120,10 +120,15 @@ class ResolveRedshiftCredentialsTransform(Transform):
         try:
             response = redshift_client.get_cluster_credentials(**params)
         except Exception as exc:
+            error_code = (
+                getattr(exc, "response", {})
+                .get("Error", {})
+                .get("Code", type(exc).__name__)
+            )
             raise CtpPipelineError(
                 stage="transform_execute",
                 step_name=step_name,
-                message=f"Failed to get Redshift cluster credentials: {exc}",
+                message=f"Failed to get Redshift cluster credentials: {error_code}",
             ) from exc
 
         return response["DbUser"], response["DbPassword"]
--- a/apollo/integrations/ctp/transforms/resolve_redshift_credentials.py
+++ b/apollo/integrations/ctp/transforms/resolve_redshift_credentials.py
@@ -34,25 +34,12 @@ class ResolveRedshiftCredentialsTransform(Transform):
       - ``password``: key in ``state.derived`` where the resolved ``DbPassword`` string is stored
     """
 
-    _REQUIRED_INPUT = ("cluster_identifier", "db_user", "db_name", "aws_region")
-    _REQUIRED_OUTPUT = ("user", "password")
+    required_input_keys = ("cluster_identifier", "db_user", "db_name", "aws_region")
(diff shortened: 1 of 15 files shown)


walles2:
## Review fixes

Addressed F2 from code review:

**F2 — `duration_seconds` type coercion** (`resolve_redshift_credentials.py`): Wrapped `int(duration_seconds)` in a try/except that raises `CtpPipelineError(stage="transform_input")` on `ValueError`/`TypeError`, keeping invalid input within the CTP error contract. Added a test (`test_invalid_duration_seconds_raises_ctp_error`).

apollo/integrations/ctp/transforms/resolve_redshift_credentials.py
```diff
@@ -0,0 +1,186 @@
...
+    Output keys:
+      - ``user``:     key in ``state.derived`` where the resolved ``DbUser`` string is stored
+      - ``password``: key in ``state.derived`` where the resolved ``DbPassword`` string is stored
+    """
+
+    _REQUIRED_INPUT = ("cluster_identifier", "db_user", "db_name", "aws_region")
+    _REQUIRED_OUTPUT = ("user", "password")
+
+    def execute(self, step: TransformStep, state: PipelineState) -> None:
+        for key in self._REQUIRED_INPUT:
```
benitezfede: Can we move these "for" to a base class and we just pass the message? They probably are the same in all the executions.

benitezfede (approved):
LGTM

PR #274, opened 2026-04-16
fix(ctp): add missing HTTP transport fields to Hive CTP

benitezfede:
## Summary

- The Hive CTP `field_map` only mapped 8 of the ~18 fields in `HiveClientArgs`, silently dropping `use_http_transport`, `http_path`, and `kerberos_service_name` when credentials passed through the pipeline
- Without `use_http_transport`, impyla defaults to binary Thrift instead of HTTP transport — breaking all HTTP-mode and Databricks-mode Spark connections routed through the agent
- Without `http_path`, the connection cannot route to the correct Spark SQL endpoint (e.g. `sql/protocolv1/o/<workspace>/<cluster>`)
- This caused queries through the agent to return empty results while the same queries worked directly from the data-collector, with misleading error messages ("Query must return at least one value")

## Root cause

The Hive CTP was registered in `90798db` (2026-04-02) with a `field_map` that only covered binary-mode connection parameters. HTTP/Databricks mode parameters (`use_http_transport`, `http_path`) and Kerberos parameters (`kerberos_service_name`) were declared in the `HiveClientArgs` TypedDict schema but not in the mapper — so the mapper produced output without them, and the schema validation didn't flag the omission since they're `NotRequired`.

## Test plan

- [x] Existing CTP tests pass (349 passed)
- [x] New test: HTTP/Databricks mode credentials survive the pipeline with all fields intact
- [x] New test: Kerberos credentials survive the pipeline
- [x] New test: missing optional fields (including new ones) are correctly omitted

### DIFF
--- a/apollo/integrations/ctp/defaults/hive.py
+++ b/apollo/integrations/ctp/defaults/hive.py
@@ -49,6 +49,9 @@ HIVE_DEFAULT_CTP = CtpConfig(
             "user": "{{ raw.user | default(none) }}",
             "password": "{{ raw.password | default(none) }}",
             "use_ssl": "{{ raw.use_ssl | default(none) }}",
+            "use_http_transport": "{{ raw.use_http_transport | default(none) }}",
+            "http_path": "{{ raw.http_path | default(none) }}",
+            "kerberos_service_name": "{{ raw.kerberos_service_name | default(none) }}",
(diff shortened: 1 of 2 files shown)


mrostan (approved):
lgtm

PR #275, opened 2026-04-22
feat: Informatica CTP support — resolve_informatica_session transform + auth-agnostic proxy client

walles2:
## Summary

- Adds `resolve_informatica_session` CTP transform supporting V2 password, V3 password, and JWT `loginOAuth` auth modes
- Refactors `InformaticaProxyClient` to be auth-method agnostic — it receives a pre-resolved `session_id` + `api_base_url` from CTP rather than performing login itself
- Registers the Informatica default CTP config (guarded by `when="raw.session_id is not defined"` so the pre-shaped DC path still works)
- Registers `informatica` in `proxy_client_factory`

## Motivation

Previously the proxy client contained login logic tied to specific auth methods. Moving session resolution into a CTP transform:
- Decouples auth from the proxy client (same pattern as other CTP connectors)
- Enables custom per-customer CTP configs (e.g. JWT via Okta ROPC → `loginOAuth`) without touching the proxy client
- Supports self-hosted credential passthrough via AWS Secrets Manager

## What's tested

- `tests/ctp/test_resolve_informatica_session.py` — 23+ tests across V2, V3, JWT, and error paths
- `tests/ctp/test_informatica_ctp.py` — pipeline integration tests
- `tests/test_informatica_proxy_client.py` — proxy client tests against the new connect_args shape

## Notes

- `INFORMATICA_PROXY_CLIENT_AGENT_MIN_VERSION` in the DC is set to `9_999_999` temporarily; will be updated to the real build number once this merges and a dev image is cut
- JWT loginOAuth requires `matchType: "aliasName"` in the Informatica IDP config and the service account user set to "IDP with SAML" authentication type

### DIFF
--- a/apollo/integrations/ctp/transforms/resolve_informatica_session.py
+++ b/apollo/integrations/ctp/transforms/resolve_informatica_session.py
@@ -192,7 +192,7 @@ class ResolveInformaticaSessionTransform(Transform):
         try:
             response = requests.post(
                 f"{base_url}{_V3_LOGIN_PATH}",
-                data={"username": username, "password": password},
+                json={"username": username, "password": password},
             )
             response.raise_for_status()
             body = response.json()
--- a/apollo/integrations/ctp/transforms/resolve_informatica_session.py
+++ b/apollo/integrations/ctp/transforms/resolve_informatica_session.py
@@ -18,12 +18,23 @@ _DEFAULT_BASE_URL = "https://dm-us.informaticacloud.com"
 def _login_failure_detail(exc: Exception) -> str:
     """Return a safe, diagnostic detail string for a login exception.
 
-    Includes HTTP status code for HTTPError (safe — response body is never
-    included), or the exception class name for other errors. Credentials are
-    never present in either form.
+    Includes HTTP status code and Informatica error text for HTTPError (safe —
+    response body never contains credentials), or the exception class name for
+    other errors.
     """
     if isinstance(exc, requests.HTTPError) and exc.response is not None:
-        return f"HTTP {exc.response.status_code}"
+        detail = f"HTTP {exc.response.status_code}"
+        try:
+            body = exc.response.json()
+            # Informatica error responses use "@type" and "message" or top-level "message"
+            message = body.get("message") or body.get("error")
+            if message:
+                detail += f": {message}"
+        except Exception:
+            text = (exc.response.text or "")[:200].strip()
+            if text:
+                detail += f": {text}"
+        return detail
     return type(exc).__name__
 
 
--- a/apollo/integrations/ctp/transforms/resolve_informatica_session.py
+++ b/apollo/integrations/ctp/transforms/resolve_informatica_session.py
@@ -1,5 +1,3 @@
-from typing import Optional
-
 import requests
 
 from apollo.integrations.ctp.errors import CtpPipelineError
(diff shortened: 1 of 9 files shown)


apollo/integrations/ctp/transforms/resolve_informatica_session.py
```diff
@@ -0,0 +1,280 @@
...
+    @staticmethod
+    def _login_v2(
+        base_url: str,
+        username: str,
+        password: str,
+        step_name: str,
+    ) -> tuple[str, str]:
+        """POST /ma/api/v2/user/login → (api_base_url, session_id)."""
+        try:
+            response = requests.post(
```
97nitt, line 170: **Missing HTTP timeout on login request**

This `requests.post` (and the same pattern at lines 204 and 253 for `_login_v3` / `_login_jwt`) doesn't pass a `timeout=`. If Informatica's login endpoint is slow, hung, or unreachable (DNS failure, firewall, half-open TCP), the agent thread will block indefinitely, holding connections and blocking other operations.

The sibling OAuth transform already uses `timeout=30` for exactly this reason — see `apollo/integrations/ctp/transforms/oauth.py:100-102`.

Suggest adding `timeout=30` to all three `requests.post` calls (or a module-level `_LOGIN_TIMEOUT = 30` constant for clarity), plus a test that patches `requests.post` and asserts `call_kwargs["timeout"] == 30`.

apollo/integrations/ctp/transforms/resolve_informatica_session.py
```diff
@@ -0,0 +1,280 @@
+from typing import Optional
```
97nitt: `Optional` is imported but never used in a type annotation — it only appears as a bare word in docstring labels. Safe to remove; ruff/flake8 will flag `F401` if they run in CI.

tests/test_informatica_proxy_client.py
```diff
@@ -0,0 +1,259 @@
...
+            self._OPERATION,
+            {"connect_args": {}},
+        )
+        self.assertIn(ATTRIBUTE_NAME_ERROR, response.result)
+        # Error must mention what's needed — not expose internal state or credentials.
+        error = response.result[ATTRIBUTE_NAME_ERROR]
+        self.assertTrue(
+            len(error) > 0,
+            "Error message must be non-empty",
+        )
```
97nitt, line 232: **Weak assertion for an "actionable error" test**

The test name and docstring promise the error is *actionable* — a user reading it knows what to do. But `len(error) > 0` is tautologically satisfied by any non-empty message, including a cryptic traceback or `"<MagicMock ...>"`.

Suggest asserting the message contains a meaningful token so a regression to an opaque error would actually fail:

```python
self.assertTrue(
    any(token in error for token in ("session_id", "connect_args", "credentials")),
    f"Error must name the missing input: {error!r}",
)
```

walles2:
Addressed all review comments in e8040b5:

- **Missing timeouts** — added `_LOGIN_TIMEOUT = 30` constant (matching OAuth transform convention) and `timeout=_LOGIN_TIMEOUT` to all three `requests.post` calls (V2, V3, JWT). Added 3 targeted tests in `TestLoginTimeout` to verify each path passes the timeout.
- **Unused `Optional` import** — removed
- **Weak assertion in `test_missing_credentials_gives_actionable_error`** — strengthened from `len(error) > 0` to checking that the error message names the missing credential key (`username`, `jwt_token`, or `session_id`), so regressions in error quality will fail the test

97nitt (approved):

PR #277, opened 2026-04-27
YET-1030: add SSL options to git HTTPS clone

genehynson:
## Summary

- Adds optional `ssl_options` (CA bundle / skip-verify) to `GitCloneClient` so the agent's git subprocess can trust a customer-supplied root CA when cloning self-managed Git providers (e.g. Flutter's self-managed GitLab via Looker Git Clone).
- Pre-existing latent bug fix: `lstrip("https://")` was eating leading `s`/`h`/`t`/`p`/`:`/`/` characters from the repo URL — replaced with `removeprefix`.

## Behavior

- `ssl_options.ca_data` (PEM): written to `/tmp/git_ca_bundle.pem`, passed to git via `-c http.sslCAInfo=<file>`.
- `ssl_options.skip_verification` / `skip_cert_verification` (alias for monolith v2 schema): when truthy, sets `-c http.sslVerify=false` and logs a warning.
- Unknown `ssl_options` keys are ignored. `ssl_options` is a no-op for SSH clones.
- Absent `ssl_options` preserves existing behavior — no risk to existing customers.

## Coordination

This is the first of three coordinated PRs (apollo-agent → data-collector → monolith-django) for [YET-1030](https://linear.app/montecarlodata/issue/YET-1030/add-ssl-options-to-the-git-agent-client). Upstream PRs:

- data-collector: bumps `GIT_AGENT_MIN_VERSION` to gate `ssl_options` send + adds matching support to legacy DC `GitCloneClientWrapper`.
- monolith-django: exposes `ssl_options` on the `TestLookerGitCloneCredentials*` and `UpdateLookerGitAuthCredentialsV2` GraphQL surfaces.

## Test plan

- [x] Unit tests added: no ssl_options, ca_data only, skip_verification, skip_cert_verification alias, both flags together, unknown keys ignored, SSH clone ignores ssl_options.
- [x] Existing `test_ssh_clone` and `test_git_version` still pass.
- [x] Black + pyright pre-commit hooks clean.
- [x] Smoke test in dev against a self-managed git provider with private CA (post-merge, manual).

### DIFF
--- a/apollo/integrations/git/git_client.py
+++ b/apollo/integrations/git/git_client.py
@@ -6,7 +6,7 @@ import shutil
 from dataclasses import dataclass
 from itertools import chain
 from pathlib import Path
-from typing import Dict, List, Generator
+from typing import Dict, List, Generator, Optional
 
 import git
 
@@ -23,11 +23,22 @@ class GitCloneClient:
     """
     Git client used to clone a repo, both ssh and https protocols are supported.
     It exposes a method that returns an iterator for all files matching a list of extensions.
+
+    Optional ``ssl_options`` (dict, HTTPS only) honored on the credentials payload:
+
+    - ``ca_data`` (str, PEM-encoded): inline CA bundle written to ``_CA_BUNDLE_FILE`` and
+      passed to git via ``-c http.sslCAInfo``. Use this for self-managed Git providers
+      whose root CA is not in the agent image's default trust store.
+    - ``skip_verification`` / ``skip_cert_verification`` (bool): when truthy, sets
+      ``-c http.sslVerify=false``. Logged as a warning.
+
+    Unknown keys in ``ssl_options`` are ignored. ``ssl_options`` is no-op for SSH clones.
     """
 
     # Only /tmp is writable in lambda.
     _REPO_DIR = Path("/tmp/repo")
     _KEY_FILE = Path("/tmp/mcd_rsa")
+    _CA_BUNDLE_FILE = Path("/tmp/git_ca_bundle.pem")
 
     GIT_TYPE = {"ssh": "SSH", "https": "HTTPS"}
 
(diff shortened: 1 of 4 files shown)


apollo/integrations/git/git_client.py
```diff
@@ -117,13 +129,51 @@ def _https_git_clone(self):
             # This allows for support of bitbucket as they handle access tokens slightly differently
             url = f"https://{self._username}:{self._token}@{self._repo_url}"
         # Use depth 1 to bring only the latest revision.
-        git_params = ["clone", "--depth", "1", url, str(self._REPO_DIR)]
+        ssl_config = self._build_ssl_config_args()
```
walles2, line 131: Git is a little funky in how we use the underlying client, but you may be able to avoid this internal function around how we transform the ssl args by using the ctp transform for resolving the ssl options (https://github.com/monte-carlo-data/apollo-agent/blob/main/apollo/integrations/ctp/transforms/resolve_ssl_options.py), but I don't know if that actually makes this easier. You could just get the ssl config as arguments in the connect args that come into the client.

genehynson: Done in e8a3989 — refactored to use the `resolve_ssl_options` transform. `GIT_DEFAULT_CTP` now runs the transform when `raw.ssl_options` is defined and the mapper exposes two scalar connect args (`ssl_ca_path`, `ssl_skip_verification`). `GitCloneClient` just reads those off `connect_args` and adds `-c http.sslCAInfo=<path>` / `-c http.sslVerify=false` to the git invocation — no more bespoke `_build_ssl_config_args` / `_write_ca_bundle`. Matches the starburst-enterprise / redshift / db2 pattern. Thanks for the pointer.

walles2 (approved):
lgtm! I know some of the ctp stuff is a little confusing.

sassmith (approved):

PR #280, opened 2026-05-01
add Informatica v2 CTP pipeline (OAuth + password auth modes)

97nitt:
## Changes

Adds the agent-side credential transform pipeline for the upcoming `informatica-v2` connection type. v2 supports **two auth modes**, discriminated by an `auth_mode` field in the credentials:

- **`auth_mode="oauth"`** — OAuth grant against the customer's IDP → JWT → Informatica `/loginOAuth`. Any grant type the shared `oauth` transform handles works (we've exercised both `client_credentials` and `password` grants against Okta in dev). Requires SAML federation configured org-wide on the customer's Informatica org.
- **`auth_mode="password"`** — direct V2 or V3 password login at Informatica (same shape v1 uses). For customers without an IDP.

The existing `InformaticaProxyClient` and `resolve_informatica_session` transform are reused as-is — both are auth-method agnostic; they only need a `session_id` + `api_base_url` at the end of the pipeline.

What this PR adds:

- **`apollo/integrations/ctp/defaults/informatica_v2.py`** — new `INFORMATICA_V2_DEFAULT_CTP` pipeline. Two transform steps, both gated on `raw.session_id is not defined` so the DC pre-shaped path bypasses them:
  1. **`oauth`** transform — runs only when `auth_mode == 'oauth'`. Performs the configured OAuth grant against the customer's IDP token endpoint and outputs a JWT. The OAuth config blob is passed through from `raw.oauth`, matching the shape monolith already stores for Snowflake.
  2. **`resolve_informatica_session`** — picks the right Informatica login path from whichever fields are present. In OAuth mode, exchanges the JWT from step 1 at `/ma/api/v2/user/loginOAuth`. In password mode, does a V2 or V3 login with `username`/`password`.
- **`apollo/integrations/ctp/registry.py`** — register the new defaults module in `_discover()`.
- **`apollo/agent/proxy_client_factory.py`** — map `"informatica-v2"` to the existing `_get_proxy_client_informatica` factory function. Same proxy client class for both v1 and v2; the only difference is which CTP pipeline resolved the session.

v1's `INFORMATICA_DEFAULT_CTP` is intentionally left untouched.

## Testing

Local unit tests:

- `tests/ctp/test_informatica_v2_ctp.py` covers pipeline registration, end-to-end OAuth happy path, password-mode happy path, IDP error surfacing as `CtpPipelineError`, Informatica `/loginOAuth` error surfacing, default base-url fallback, and absence of raw credentials in the resolved output.
- `tests/test_informatica_proxy_client.py::InformaticaV2ProxyClientReuseTests` confirms `"informatica-v2"` drives the same `do_http_request` path as v1.
- v1 informatica suites unchanged: `tests/ctp/test_informatica_ctp.py`, `tests/ctp/test_resolve_informatica_session.py`, and the rest of `tests/test_informatica_proxy_client.py` all pass without modification.

```
$ pytest tests/ctp/test_informatica_v2_ctp.py tests/ctp/test_informatica_ctp.py \
         tests/ctp/test_resolve_informatica_session.py tests/test_informatica_proxy_client.py
======================== 52 passed in 1.68s ========================
```

`black` and `pyright` are clean on changed files.

End-to-end verification in dev:

- **OAuth mode (password grant against Okta IDP)** — verified successful authentication and full extractor walk against the Informatica sandbox via the **data-collector** direct path, the **apollo-agent** path, and the **hermes-agent** path (file-mounted self-hosted credentials).
- **Password mode (direct V3 login at Informatica)** — verified successful authentication and full extractor walk against the same sandbox via the **data-collector** direct path, the **apollo-agent** path, and the **hermes-agent** path.

In each case the validator's `validate_informatica_v2_api_calls` check returned `success: true` with the expected per-call `additionalData.queriesWithResults` entries.

## Context

Part of the broader Informatica v2 (mapping-task-first) integration. v1's discovery model (taskflow-first) misses mapping tasks customers run outside any taskflow; v2 pivots to mapping-task-first discovery via the V2 `/mttask` listing endpoint.

Authentication is the other big change. v1 is username/password only. v2 supports both OAuth (for customers with an IDP and SAML-federated Informatica orgs) and password (for customers without an IDP). The CTP pipeline branches on `auth_mode` rather than committing to a single auth flow — keeps v2 usable for customers across the spectrum without forcing a separate v3 down the road.

Coordinated with monolith-django#13339 (Phase 2: monolith connection enablement) and the upcoming data-collector PR (Phase 3: collection-side validator + extractor + OAuth/password auth classes). Once this PR merges, an apollo-agent release should be tagged so the data-collector can pin its `INFORMATICA_V2_AGENT_MIN_VERSION` to a known-good version.

### DIFF
--- a/apollo/integrations/ctp/defaults/informatica_v2.py
+++ b/apollo/integrations/ctp/defaults/informatica_v2.py
@@ -10,31 +10,35 @@ class InformaticaV2ClientArgs(TypedDict):
     api_base_url: Required[str]
 
 
-# Informatica v2 authenticates via OAuth → JWT → /loginOAuth, using the same
-# `OAuthConfiguration` shape monolith stores for Snowflake. v1 (username/password)
-# keeps using INFORMATICA_DEFAULT_CTP.
+# v2 supports two auth modes, discriminated by `raw.auth_mode`:
+#   - "oauth"    → OAuth grant against the customer's IDP → JWT → /loginOAuth
+#   - "password" → V2 or V3 password login (same as v1's INFORMATICA_DEFAULT_CTP)
+# `resolve_informatica_session` picks the right Informatica login path from
+# whichever fields are present. We only need a leading OAuth step in oauth mode.
 INFORMATICA_V2_DEFAULT_CTP = CtpConfig(
     name="informatica-v2-default",
     steps=[
-        # Step 1: OAuth grant against the customer's IDP. Whatever grant_type
-        # `raw.oauth` declares (client_credentials, password, etc.) is what the
-        # shared `oauth` transform performs. Skipped when the session is already
-        # pre-resolved.
+        # Step 1 (OAuth mode only): OAuth grant against the customer's IDP using
+        # the shared `oauth` transform. Whatever grant_type `raw.oauth` declares
+        # (client_credentials, password, etc.) is what the transform performs.
         TransformStep(
             type="oauth",
-            when="raw.session_id is not defined",
+            when="raw.session_id is not defined and raw.auth_mode == 'oauth'",
             input={"oauth": "{{ raw.oauth }}"},
             output={"token": "informatica_jwt"},
         ),
-        # Step 2: Exchange the JWT at /ma/api/v2/user/loginOAuth for an
-        # icSessionId + API base URL. Skipped when pre-resolved.
+        # Step 2: Resolve the Informatica session. In oauth mode, exchange the
+        # JWT from step 1 at /loginOAuth. In password mode, do a V2 or V3 login.
         TransformStep(
             type="resolve_informatica_session",
             when="raw.session_id is not defined",
             input={
-                "jwt_token": "{{ derived.informatica_jwt }}",
-                "org_id": "{{ raw.org_id }}",
+                "username": "{{ raw.username | default(none) }}",
+                "password": "{{ raw.password | default(none) }}",
+                "informatica_auth": "{{ raw.informatica_auth | default(none) }}",
                 "base_url": "{{ raw.base_url | default(none) }}",
+                "jwt_token": "{{ derived.informatica_jwt | default(none) }}",
+                "org_id": "{{ raw.org_id | default(none) }}",
             },
             output={
                 "session_id": "informatica_session_id",
--- a/apollo/integrations/ctp/defaults/informatica_v2.py
+++ b/apollo/integrations/ctp/defaults/informatica_v2.py
@@ -10,36 +10,24 @@ class InformaticaV2ClientArgs(TypedDict):
     api_base_url: Required[str]
 
 
-# Informatica v2 authenticates via OAuth client_credentials → JWT → /loginOAuth.
-# v1 (username/password) keeps using INFORMATICA_DEFAULT_CTP; this pipeline is
-# separate so the two auth flows don't share validation surface area.
+# Informatica v2 authenticates via OAuth → JWT → /loginOAuth, using the same
+# `OAuthConfiguration` shape monolith stores for Snowflake. v1 (username/password)
+# keeps using INFORMATICA_DEFAULT_CTP.
 INFORMATICA_V2_DEFAULT_CTP = CtpConfig(
     name="informatica-v2-default",
     steps=[
-        # Step 1: OAuth client_credentials grant against the customer's IDP token
-        # endpoint. The shared `oauth` transform expects an `oauth` config dict;
-        # build it inline from the flat raw credential fields. Skipped when the
-        # session is already pre-resolved (DC pre-shaped path or custom CTP
-        # config that resolved upstream).
+        # Step 1: OAuth grant against the customer's IDP. Whatever grant_type
+        # `raw.oauth` declares (client_credentials, password, etc.) is what the
+        # shared `oauth` transform performs. Skipped when the session is already
+        # pre-resolved.
         TransformStep(
             type="oauth",
             when="raw.session_id is not defined",
-            input={
-                "oauth": (
-                    "{{ {"
-                    "'client_id': raw.client_id, "
-                    "'client_secret': raw.client_secret, "
-                    "'access_token_endpoint': raw.token_url, "
-                    "'grant_type': 'client_credentials'"
-                    "} }}"
-                ),
-            },
+            input={"oauth": "{{ raw.oauth }}"},
             output={"token": "informatica_jwt"},
         ),
-        # Step 2: Exchange the JWT at Informatica's /ma/api/v2/user/loginOAuth
-        # endpoint for an icSessionId + API base URL. Reuses the v1 transform's
-        # JWT mode (requires SAML federation configured org-wide). Skipped when
-        # the session is already pre-resolved.
+        # Step 2: Exchange the JWT at /ma/api/v2/user/loginOAuth for an
+        # icSessionId + API base URL. Skipped when pre-resolved.
         TransformStep(
             type="resolve_informatica_session",
             when="raw.session_id is not defined",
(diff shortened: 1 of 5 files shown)


mrostan (commented):
Took a closer look — overall this looks great. Auth-mode discriminator pattern is well-chosen, the `when` gates are right, and reusing v1's proxy client via a distinct `connection_type` is the right factoring. Just a handful of nits, none of them blocking.

tests/ctp/test_informatica_v2_ctp.py
```diff
@@ -0,0 +1,255 @@
...
+_PASSWORD_MODE_CREDENTIALS = {
+    "auth_mode": "password",
+    "username": "svc-user",
+    "password": "svc-pass",
+    "informatica_auth": "v3",
+    "base_url": "https://dm-us.informaticacloud.com",
+}
+
+# Backwards-compat alias for tests written before the auth_mode discriminator.
+_FLAT_CREDENTIALS = _OAUTH_MODE_CREDENTIALS
```
mrostan: nit: this comment doesn't quite fit a brand-new file — there are no "tests written before the auth_mode discriminator" here. Looks like a copy-paste artifact from a draft. Either drop the alias and use `_OAUTH_MODE_CREDENTIALS` directly at the call sites, or keep the alias with a comment that reflects its actual purpose.

apollo/integrations/ctp/defaults/informatica_v2.py
```diff
@@ -0,0 +1,64 @@
...
+#   - "oauth"    → OAuth grant against the customer's IDP → JWT → /loginOAuth
+#   - "password" → V2 or V3 password login (same as v1's INFORMATICA_DEFAULT_CTP)
+# `resolve_informatica_session` picks the right Informatica login path from
+# whichever fields are present. We only need a leading OAuth step in oauth mode.
+INFORMATICA_V2_DEFAULT_CTP = CtpConfig(
+    name="informatica-v2-default",
+    steps=[
+        # Step 1 (OAuth mode only): OAuth grant against the customer's IDP using
+        # the shared `oauth` transform. Whatever grant_type `raw.oauth` declares
+        # (client_credentials, password, etc.) is what the transform performs.
```
mrostan: nit: the shared `oauth` transform validates against `_SUPPORTED_GRANT_TYPES = ("client_credentials", "password")` and raises for anything else — "etc." here implies more flexibility than there is. "client_credentials or password" would be accurate.

tests/ctp/test_informatica_v2_ctp.py
```diff
@@ -0,0 +1,255 @@
...
+        ]
+
+        creds_without_base_url = {
+            k: v for k, v in _FLAT_CREDENTIALS.items() if k != "base_url"
+        }
+        CtpPipeline().execute(INFORMATICA_V2_DEFAULT_CTP, creds_without_base_url)
+
+        # Second call must still go to a default Informatica login URL
+        second_call_url = mock_post.call_args_list[1][0][0]
+        self.assertIn("informaticacloud.com/ma/api/v2/user/loginOAuth", second_call_url)
```
mrostan: nit: this substring assertion passes for any `*.informaticacloud.com` POD URL. The test name promises the *default* is used, but if the default in `resolve_informatica_session` silently drifts (dm-us → dm-eu, or to a staging URL), nothing here catches it. Pinning the specific hostname (or asserting exact equality with `f"{_DEFAULT_BASE_URL}{_LOGIN_OAUTH_PATH}"`) would make the test do what its name promises.

tests/ctp/test_informatica_v2_ctp.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+def _mock_post(body: dict) -> MagicMock:
+    resp = MagicMock()
+    resp.json.return_value = body
+    resp.raise_for_status.return_value = None
+    return resp
+
+
+class TestInformaticaV2CtpRegistered(TestCase):
```
mrostan, line 57: nit: v1's suite has `TestInformaticaCtpPreResolvedPath` that exercises the DC pre-shaped path (`{"session_id": "...", "api_base_url": "..."}` → both steps skipped, `mock_post.assert_not_called()`). v2 has the same `when="raw.session_id is not defined"` guards on both steps but no equivalent test, so a future change that breaks the guard would only surface in production. Worth adding the parallel test for symmetry.

tests/ctp/test_informatica_v2_ctp.py
```diff
@@ -0,0 +1,255 @@
...
+}
+
+_OAUTH_MODE_CREDENTIALS = {
+    "auth_mode": "oauth",
+    "oauth": _OAUTH_CONFIG_CLIENT_CREDENTIALS,
+    "org_id": "ORG-12345",
+    "base_url": "https://dm-us.informaticacloud.com",
+}
+
+_PASSWORD_MODE_CREDENTIALS = {
```
mrostan, line 41: nit: `auth_mode` has no negative coverage. `auth_mode` missing entirely (legacy creds), `"OAuth"` (case mismatch), `None`, `"invalid"` — all silently fall through to the password path, because step 1's `when` evaluates `raw.auth_mode == 'oauth'` to `False` for any of them. The behavior is probably what you want (it surfaces as a clear error from `_require("username", ...)`), but a single test pinning the case-mismatch path would make the intent explicit and prevent a future refactor from accidentally making the discriminator case-insensitive.

mrostan (approved):
lgtm, shared a few nits by Claude, feel free to ignore

tests/ctp/test_informatica_v2_ctp.py
97nitt: Fixed in d09cfd3 — dropped the `_FLAT_CREDENTIALS` alias and updated all 9 call sites to use `_OAUTH_MODE_CREDENTIALS` directly. Right read on the copy-paste — it was a holdover from before the discriminator landed.

apollo/integrations/ctp/defaults/informatica_v2.py
97nitt: Fixed in e84970c — comment now says "client_credentials or password — the only types the shared transform supports." Agreed, "etc." overpromised vs. what `_SUPPORTED_GRANT_TYPES` actually allows.

tests/ctp/test_informatica_v2_ctp.py
97nitt: Fixed in eff12fd — imported `_DEFAULT_BASE_URL` from `resolve_informatica_session` and asserted exact equality with `f"{_INFORMATICA_DEFAULT_BASE_URL}/ma/api/v2/user/loginOAuth"`. Now any drift in the default URL fails the test.

tests/ctp/test_informatica_v2_ctp.py
97nitt: Fixed in 852c413 — added `TestInformaticaV2CtpPreResolvedPath` mirroring v1's. Asserts `mock_post.assert_not_called()` and that the pre-shaped `session_id` / `api_base_url` flow through unchanged when both `when=` guards trigger.

tests/ctp/test_informatica_v2_ctp.py
97nitt: Fixed in f103970 — added `TestInformaticaV2CtpAuthModeDiscriminator::test_case_mismatch_auth_mode_falls_through_to_password_path`. Pins the "strict equality" intent: `auth_mode: "OAuth"` (uppercase) skips step 1 (verified via `mock_post.assert_not_called()`) and surfaces as a `CtpPipelineError` from password-mode field validation. A future refactor that makes the discriminator case-insensitive (or coerces missing values) will fail this test.

PR #282, opened 2026-05-05
ci: split system-base from venv-included base, publish system-base

mrostan:
## Summary
- Split the Dockerfile's monolithic `base` stage into a new `system-base` (apt-only) plus the existing `base` (venv + pip install + apollo source) that now extends it.
- Replace the `generic-base` publish target with `system-base` (`latest-system-base` + `<version>-system-base`).
- Apollo's own targets (`cloudrun`, `azure`, `aws_proxied`, `aws_generic`, `tests`, `lambda`) are functionally unchanged — they still `FROM base` and produce identical images.

## Why
The previous `generic-base` image bundled apollo's full venv on top of the apt layer. Downstream consumers (hermes-agent specifically) inherited that venv, then ran their own `pip install` with their own pinned versions. Where pins diverged (~50 transitive deps in hermes' case), pip reinstalled — adding ~340 MB of duplicate package files on top of an already-installed copy in the lower layer. Net: hermes images were ~1 GB compressed when they should have been ~600 MB.

The fix is structural: stop trying to share the venv. `system-base` shares only the apt/native lib layer (where sharing actually works because there's no version drift), and downstream consumers build a clean venv on top with exactly the deps they want.

## Backwards compatibility
- `latest-generic-base` and `<version>-generic-base` (introduced in #281, present in the v1.5.3 release) stop being published from this PR forward. The existing v1.5.3 tags remain on Docker Hub for any consumer pinned to them.
- Hermes-agent currently consumes `1.5.3-generic-base`. A follow-up hermes PR will switch to `<next-version>-system-base` once this lands and a release is cut.
- No other known consumers of `generic-base`.

## Test plan
- [x] `docker build --target system-base` builds cleanly; `/app/` is empty (no apollo source, no venv); image size ~300 MB uncompressed
- [x] `docker build --target base` builds cleanly with identical end-state to today (apt + venv + apollo source)
- [x] `docker build --progress=plain --target base` shows 5 wheel-build entries (\`agent-base\`, \`oscrypto\`, \`pure-sasl\`, \`thrift\`, \`google-crc32c\`) — all pure-Python; no \`gcc\`/\`cc1\`/header-error lines, confirming no C compilation needs apt-installed headers and the new ordering (all apt before venv/pip) is safe
- [ ] CircleCI: build pipeline succeeds end-to-end including version-verify on the four apollo targets and the docker push of all tags including the new \`*-system-base\` ones
- [ ] After merge + release: confirm \`montecarlodata/agent:latest-system-base\` and \`montecarlodata/agent:<version>-system-base\` exist on Docker Hub
- [ ] Hermes follow-up PR consumes the new tag and reports a smaller image

## Follow-up
Hermes-agent PR #37 will be updated to:
- Swap \`FROM montecarlodata/agent:1.5.3-generic-base\` → \`FROM montecarlodata/agent:<next-version>-system-base\`
- Add \`python -m venv\` (no longer inherited)
- Drop the \`rm -rf ./apollo\` workaround (no apollo source in system-base)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -94,15 +94,18 @@ jobs:
       - checkout
       - docker/check:
           use-docker-credentials-store: true
-      # Publish the `base` stage as `<version>-generic-base` so downstream
-      # repos (e.g. hermes-agent) can pin to a specific apollo version that
-      # matches their pip-installed apollo dependency.
+      # Publish the `system-base` stage (apt-only, no venv, no apollo source) as
+      # `<version>-system-base`. Downstream consumers (e.g. hermes-agent) build
+      # their own venv on top against the same native libs without inheriting
+      # apollo's pip-installed Python deps. Replaces the previous `generic-base`
+      # tag, which bundled the venv and caused image bloat for downstream repos
+      # whose pinned versions diverged from apollo's resolved set.
       - docker/build:
-          step-name: Build generic base image
+          step-name: Build system base image
           use-buildkit: true # to build only the required stages
-          extra_build_args: --target base --build-arg code_version=<< parameters.code_version >> --build-arg build_number=<< pipeline.number >>
+          extra_build_args: --target system-base --build-arg code_version=<< parameters.code_version >> --build-arg build_number=<< pipeline.number >>
           image: montecarlodata/<< parameters.docker_hub_repository >>
-          tag: latest-generic-base,<< parameters.code_version >>-generic-base
+          tag: latest-system-base,<< parameters.code_version >>-system-base
       - docker/build:
           step-name: Build AWS generic image
           use-buildkit: true  # to build only the required stages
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -121,9 +124,8 @@ jobs:
           extra_build_args: --target azure --build-arg code_version=<< parameters.code_version >> --build-arg build_number=<< pipeline.number >>
           image: montecarlodata/<< parameters.docker_hub_repository >>
           tag: latest-azure,<< parameters.code_version >>-azure
-      - verify-version-in-docker-image:
-          image: montecarlodata/<< parameters.docker_hub_repository >>:latest-generic-base
-          version: << parameters.code_version >>,<< pipeline.number >>
+      # system-base has no apollo source, so there is no /app/apollo/agent/version
+      # file to verify — skip the version check for it.
```
97nitt: nit: maybe omit the comment above?

Dockerfile
```diff
@@ -1,17 +1,12 @@
...
-# Allow statements and log messages to immediately appear in the logs
-ENV PYTHONUNBUFFERED=True
+# system-base — system-level dependencies only (apt packages, no venv).
+# Published as `<version>-system-base` so downstream consumers (e.g. hermes-agent)
+# can build their own venv against the same native libs without inheriting
+# apollo's pip-installed dependencies.
+FROM python:3.12-slim AS system-base
 
 ENV APP_HOME=/app
 ENV VENV_DIR=.venv
```
97nitt: `ENV VENV_DIR=.venv` lives in `system-base` even though that stage installs no venv — `VENV_DIR` is only consumed by the `base` stage and its descendants. Setting it here bakes it into the published `system-base` image's environment, where it has no meaning for downstream consumers (and may surface confusingly in `docker inspect` for an agent that picks a different venv path).

Suggest moving it down to the `base` stage alongside `RUN python -m venv $VENV_DIR`. `APP_HOME` and `WORKDIR` are fine to keep at this layer — they have generic value.

97nitt (approved):
lgtm. two minor nits, one from Claude one from me.

.circleci/config.yml
mrostan: yup, good call

Dockerfile
mrostan: makes sense too, thanks

PR #286, opened 2026-05-08
fix(sup-373): preserve outer credential siblings when CtpRegistry unwraps connect_args

nobeirn:
## Summary

`CtpRegistry.resolve` / `resolve_custom` were unwrapping `connect_args` from the incoming credentials dict and running the CTP pipeline with **only** the inner dict — silently discarding any sibling fields. The data-collector's mysql agent path (`clients/plugins/plugin_mysql.py:48-51`) sends `ssl_options` as a **sibling** of `connect_args`, not nested inside it:

```python
agent_credentials = {
    "connect_args": {host, port, user, password},
    "ssl_options": {"ca_data": "<inline PEM>"},  # sibling — got dropped
}
```

After the unwrap, the pipeline ran on `{host, port, user, password}` only. The mysql CTP's `when="raw.ssl_options is defined"` clause was always False, the SSL transform never fired, `connect_args["ssl"]` was never set, and `pymysql.connect` ran without TLS. MySQL with `require_secure_transport=ON` rejected with `(3159, '... --require_secure_transport=ON.')`.

This is the root cause of [SUP-373](https://linear.app/montecarlodata/issue/SUP-373) — PennyMac's MySQL agent has been failing every 12h since 2026-04-30. Same root cause as [YET-321](https://linear.app/montecarlodata/issue/YET-321) (DB2 RDS SSL via agent), which closes transitively.

## Fix

New `_build_pipeline_input` helper in `apollo/integrations/ctp/registry.py` merges sibling fields with the inner `connect_args` dict before pipeline execution. Inner `connect_args` keys win on collision (driver-direct values are most specific). Non-dict `connect_args` (legacy pre-built ODBC strings) pass through unchanged — the proxy client's own ssl_options handling still works for that path. Both `resolve` and `resolve_custom` go through the helper, so they stay symmetric.

## Blast radius

This change is at the registry layer, so every CTP-registered integration runs through it. Three reasons it's safe:

1. **Inner wins on collision.** `{**outer_siblings, **inner}` means any field already in `connect_args` keeps its existing pipeline-visible value. No change for fields the pipeline already saw.
2. **Legacy ODBC-string path preserved exactly.** When `connect_args` is a non-dict, `_build_pipeline_input` returns `None` and `resolve` returns the original credentials dict unchanged — matching the pre-fix behavior the db2 proxy client (and others) rely on for top-level `ssl_options` handling.
3. **All 444 CTP tests pass** with no regressions across mysql, postgres, snowflake, databricks, informatica (v1 + v2), presto, power_bi, redshift, starburst, db2, oracle, etc.

**Positive blast radius:** the same unwrap-discard pattern was silently breaking ssl_options for every transactional-DB CTP that uses the `when="raw.ssl_options is defined"` pattern (postgres, redshift, starburst-enterprise, starburst-galaxy, db2, http, git). They share the registry-level path; they get fixed too.

**Theoretical risk worth a reviewer's eye:** other `when` clauses across CTPs (`raw.session_id is not defined`, `raw.auth_mode == 'oauth'`, `raw.auth is defined`, etc.) could in principle flip behavior if any DC plugin sends those fields as outer siblings rather than inside `connect_args`. The 444 existing tests cover the realistic shapes for each CTP — but I haven't audited every DC plugin's payload shape against every CTP's `when` clauses. Would value a sanity check from someone who has end-to-end visibility into DC payload shapes for non-mysql CTPs.

## Test plan

- [x] New repro test: `tests/ctp/test_mysql_ctp.py::TestMysqlCtp::test_resolve_mysql_dc_pre_shaped_credentials_with_outer_ssl_options` — exercises `CtpRegistry.resolve("mysql", ...)` with the realistic DC pre-shaped payload, asserts `connect_args["ssl"]` is an `ssl.SSLContext`. Failed before the fix; passes after.
- [x] Full CTP suite: `pytest tests/ctp/` → 444 passed, no regressions.
- [x] Black + pyright clean on changed files.
- [ ] Post-merge: verify PennyMac (account `959c52ae-f98b-47b6-b51c-ff40cdc2224b`) — once the apollo build with this fix deploys, the 12h `(3159, ...)` failure cycle on connection 128437 should stop. No customer action needed; their persisted credentials are correctly shaped.

## Coordination

Out of scope for this PR but locally on `nobeirn/sup-373-ssl-ca-nesting` in `monolith-django` and `data-collector` are diagnostic/regression tests that pin the working contract at each layer (monolith persistence and DC plugin forwarding both produce the right shape — the bug is purely apollo-side). Holding those locally for now to keep this PR focused on the fix; will follow up separately if we want the cross-layer regression coverage in CI.

## Context

Linear ticket has the full 8-step end-to-end credential trace, Datadog evidence for PennyMac's failure cycle, customer DB record details, and the affected-scope table covering mysql, oracle, mariadb, db2, and starburst-enterprise: https://linear.app/montecarlodata/issue/SUP-373

Original ticket pinned the bug to CLI-side nesting of `ssl_options` under `connection_settings`. That hypothesis was empirically refuted — monolith's GraphQL schema for transactional DBs requires the nesting, and a CLI-only fix would have regressed production. The actual bug is downstream of monolith persistence, in apollo's CTP registry — what this PR addresses.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/apollo/integrations/ctp/registry.py
+++ b/apollo/integrations/ctp/registry.py
@@ -88,12 +88,12 @@ class CtpRegistry:
                 **registered.connect_args_defaults,
                 **config.connect_args_defaults,
             }
-        raw_or_connect_args = credentials.get(_ATTR_CONNECT_ARGS, credentials)
-        if not isinstance(raw_or_connect_args, dict):
+        pipeline_input = _build_pipeline_input(credentials)
+        if pipeline_input is None:
             return credentials
         return {
             _ATTR_CONNECT_ARGS: CtpPipeline().execute(
-                config, raw_or_connect_args, context=context or {}
+                config, pipeline_input, context=context or {}
             )
         }
 
@@ -119,15 +119,36 @@ class CtpRegistry:
                 stage="registry",
                 message=f"No CTP config registered for '{connection_type}'. Call CtpRegistry.get() before resolve().",
             )
-        # Unwrap pre-shaped connect_args so both flat and DC-pre-shaped credentials
-        # follow the same transform path through the pipeline.
-        # If connect_args is not a dict (e.g. a pre-built ODBC string), pass through
-        # unchanged — the pipeline cannot interpret non-dict credentials.
-        raw_or_connect_args = credentials.get(_ATTR_CONNECT_ARGS, credentials)
-        if not isinstance(raw_or_connect_args, dict):
+        pipeline_input = _build_pipeline_input(credentials)
+        if pipeline_input is None:
             return credentials
         return {
             _ATTR_CONNECT_ARGS: CtpPipeline().execute(
-                config, raw_or_connect_args, context=context or {}
+                config, pipeline_input, context=context or {}
             )
         }
+
+
+def _build_pipeline_input(credentials: dict[str, Any]) -> dict[str, Any] | None:
+    """Return the dict to feed the CTP pipeline, or None if credentials cannot
+    be processed (e.g. pre-built ODBC string).
+
+    For DC pre-shaped credentials (a ``connect_args`` dict alongside other
+    credential fields like ``ssl_options``), merge the outer fields with the
+    inner ``connect_args`` so the pipeline sees the full credential picture.
+    Inner ``connect_args`` keys take precedence — they're driver-direct and
+    have been explicitly named.
+
+    SUP-373: without this merge, the data-collector's mysql agent path
+    (clients/plugins/plugin_mysql.py:48-51) sends ``ssl_options`` as a
+    sibling of ``connect_args``; the previous unwrap discarded it, no SSL
+    context was built, and pymysql connected without TLS — MySQL with
+    ``require_secure_transport=ON`` rejected with error 3159.
+    """
+    inner = credentials.get(_ATTR_CONNECT_ARGS)
+    if inner is None:
+        return credentials
+    if not isinstance(inner, dict):
+        return None
+    outer_siblings = {k: v for k, v in credentials.items() if k != _ATTR_CONNECT_ARGS}
+    return {**outer_siblings, **inner}
--- a/tests/ctp/test_mysql_ctp.py
+++ b/tests/ctp/test_mysql_ctp.py
@@ -131,3 +131,11 @@ class TestMysqlCtp(TestCase):
             ssl_module.SSLContext,
             f"ssl is set but not an SSLContext: {connect_args['ssl']!r}",
         )
+
+        # F8 follow-up: explicitly verify inner connect_args fields survive the merge.
+        # Without these assertions a regression that drops the inner dict would still
+        # pass the SSL check above (ssl_options would still merge in from outer).
+        self.assertEqual("db.example.com", connect_args["host"])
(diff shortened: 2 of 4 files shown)


genehynson:
## Test coverage suggestions (follow-up)

Surfaced from `/code-review` on c74a2f1. None block merge — flagging as low-cost regression coverage that would lock in the new contract.

| | Suggestion |
|---|---|
| **F5** | Pin the inner-wins precedence rule. `_build_pipeline_input` returns `{**outer_siblings, **inner}`; a future flip to `{**inner, **outer_siblings}` would pass the suite today. Suggest a test with overlapping `host` (outer=A, inner=B) → assert `connect_args["host"] == "B"`. |
| **F6** | `resolve_custom` was migrated to the same helper in this PR but `TestCustomCtpSchemaInjection` has no sibling-merge test. Add an analogous `resolve_custom` call with `{connect_args, ssl_options}` to lock symmetry with `resolve`. |
| **F7** | Non-dict `connect_args` passthrough (the `pipeline_input is None` branch) is only exercised through `resolve_custom` (`test_non_dict_connect_args_returned_unchanged`). Mirror it for `resolve`. |
| **F8** | The new `test_resolve_mysql_dc_pre_shaped_credentials_with_outer_ssl_options` asserts the SSL context but not that `host`/`port`/`user`/`password` survive the merge. A regression dropping the inner dict would still pass. |
| **F9** | Edge cases for `connect_args=None` and `connect_args={}` are not exercised. The `None` branch routes through the pipeline post-PR (was passthrough pre-PR — see also the `connect_args=None` semantic shift noted in the review). The `{}` branch now feeds outer-only to the pipeline (was empty-dict pre-PR). |

All five are SUGGESTION severity. Take or leave per your call.

genehynson (approved):

nobeirn:
Follow-up tests pushed in `1fde055` addressing the four gaps from `/code-review`:

- **F5** — `TestSiblingMergePrecedence::test_inner_connect_args_wins_on_collision` (`tests/ctp/test_registry.py`). Pins the inner-wins rule with overlapping `host`/`port`/`user`/`password`; would catch a future flip to `{**inner, **outer}`.
- **F6** — `TestCustomCtpSchemaInjection::test_outer_siblings_merged_with_inner_connect_args` (`tests/test_custom_ctp.py`). Mirrors the sibling-merge through `resolve_custom` to lock symmetry with `resolve`.
- **F8** — extended `test_resolve_mysql_dc_pre_shaped_credentials_with_outer_ssl_options` with `host`/`port`/`user`/`password` assertions so a regression dropping the inner dict can't pass on the SSL check alone.
- **F9** — `TestEmptyAndNoneConnectArgs` (2 tests) for `connect_args=None` and `connect_args={}` edge cases. Locks in the new "fall back to outer/flat fields and route through the pipeline" behavior.

**F7** was already covered by the existing `TestNonDictConnectArgsPassthrough` (4 tests exercising `resolve` with non-dict `connect_args`); leaving as-is.

466 ctp + custom_ctp tests pass; black + pyright clean.

PR #291, opened 2026-05-15
YET-1256: Add list_dataspaces SOQL operation to Salesforce Data Cloud proxy client

walles2:
## Summary

Adds a `list_dataspaces() -> list[str]` method to `SalesforceDataCloudProxyClient` so the data-collector can dispatch SOQL dataspace discovery through the agent. This is the apollo-agent companion to [YET-1177](https://linear.app/montecarloai/issue/YET-1177) (DC-side auto-discovery for non-self-hosted credentials); it unblocks [YET-1257](https://linear.app/montecarloai/issue/YET-1257), which wires the DC's self-hosted branch through this operation so self-hosted customers can also avoid manually configuring `dataspaces`.

The method mints a core OAuth token via the client-credentials grant against `/services/oauth2/token` (the `salesforcecdpconnector` library does not expose its token-minting flow for arbitrary SOQL), then issues `GET /services/data/v62.0/query?q=SELECT DataSpaceApiName FROM Dataspace` with the core token as a Bearer credential. `nextRecordsUrl` pagination is handled defensively. The HTTP status is embedded in the `code NNN` format on any non-200 response so the data-collector's `_extract_exchange_http_status` can surface it for structured Datadog queries.

Both HTTP calls flow through a single `_CapturingSession`, so the redacted response body is surfaced in the exception message and structured log record on failure — no raw `access_token` strings can leak into errors or logs.

## Companion work (same Linear ticket)

A small data-collector PR will follow under YET-1256: add an `@agent_operation`-decorated `list_dataspaces` passthrough on `SalesforceDataCloudProxyConnection` so the DC can dispatch this new method. That's a fresh `/start-work` in `data-collector` once this PR is open and the method shape is locked.

## Key Decisions

- **OAuth is done by us, not the library.** The `salesforcecdpconnector` library owns OAuth for its Data Cloud endpoints but doesn't expose token minting for arbitrary SOQL against the core REST API. We POST to `/services/oauth2/token` directly — same shape as the data-collector's `_discover_dataspaces`.
- **No retry on either HTTP call.** Parity with `_discover_dataspaces` on the DC side. The cursor's `_RETRY_KWARGS` stays scoped to query execution where stale pooled connections are the actual failure mode.
- **`_CapturingSession` generalized to capture all GET/POST traffic** (was: only `a360/token` POSTs). New `get()` override, a private `_capture` method, and dropped URL filter. Library-internal callers still work — `last_exchange_*` now reflects whichever request the library last issued before raising, which is the call that caused the error. Strict improvement for error enrichment.
- **No version-bumping in source.** Agent version is generated by CI from git tags + pipeline number into `apollo/agent/version` at Docker build time. YET-1257's `SALESFORCE_DATA_CLOUD_AGENT_MIN_VERSION` bump on the DC side will gate on whatever version emerges from this release.
- **No `@agent_operation`-equivalent registration on the agent side** — confirmed via the proxy-client factory dispatch mechanism. The agent dispatches proxy-client methods by name lookup, so adding the method is sufficient.

## Testing

- 6 new tests in `tests/test_salesforce_data_cloud_client.py`: success, empty, pagination via `nextRecordsUrl`, SOQL HTTP 401 with status embedded, OAuth HTTP 401 with body redaction (verified no token leak), HTTP 429 with `exchange_error_type=rate_limited` in the structured log record.
- 22 existing tests continue to pass — the `_CapturingSession` generalization is benign for the existing `list_tables` flow.
- Black + pyright clean on changed files.

## Linked

- Predecessor: [YET-1177](https://linear.app/montecarloai/issue/YET-1177) — DC-side auto-discovery (non-self-hosted, shipped)
- This ticket: [YET-1256](https://linear.app/montecarloai/issue/YET-1256)
- Successor: [YET-1257](https://linear.app/montecarloai/issue/YET-1257) — DC self-hosted wiring + `SALESFORCE_DATA_CLOUD_AGENT_MIN_VERSION` bump (blocked on this PR)

### DIFF
--- a/apollo/integrations/db/salesforce_data_cloud_proxy_client.py
+++ b/apollo/integrations/db/salesforce_data_cloud_proxy_client.py
@@ -45,19 +45,38 @@ _RETRY_KWARGS: dict[str, Any] = {
     "logger": logger,
 }
 
+# Salesforce core REST API version used by the SOQL dataspace discovery call
+# (see ``SalesforceDataCloudProxyClient.list_dataspaces``). Validated against
+# customer orgs returning DataSpaceApiName from the Dataspace SObject (YET-1256).
+_SOQL_API_VERSION = "v62.0"
+_LIST_DATASPACES_SOQL = "SELECT DataSpaceApiName FROM Dataspace"
+_DISCOVERY_REQUEST_TIMEOUT_SECONDS = 30
+
 
 class _CapturingSession(requests.Session):
     """
-    Wraps the library's requests.Session to capture the Salesforce a360/token response
-    body regardless of status code.
+    Wraps requests.Session to capture the body and status of every HTTP call
+    made through this session, regardless of method, URL, or status code.
 
-    The salesforcecdpconnector library raises Error('CDP token retrieval failed with
-    code N') on non-200 and discards the body. On 200 responses with unexpected payloads
-    (e.g. Salesforce returning 200 with an error body for unknown dataspaces) the library
-    raises KeyError. In both cases the body is discarded before we can see it.
+    Two use patterns:
 
-    Capturing on all a360/token responses means we always have it available to include
-    in the RuntimeError that propagates back to the data-collector and into Datadog logs.
+    1. **Library-internal capture.** Caller swaps this session into the
+       salesforcecdpconnector library's authentication_helper via
+       `_attach_capturing_session` so the library's internal HTTP traffic is
+       captured. The library raises ``Error('CDP token retrieval failed with
+       code N')`` on non-200 a360/token responses and discards the body; on
+       200 responses with unexpected payloads it raises ``KeyError`` after
+       discarding the body. ``last_exchange_*`` reflects whichever request the
+       library last issued before raising, which is the call that caused the
+       error.
+    2. **Direct capture.** Caller instantiates and uses the session directly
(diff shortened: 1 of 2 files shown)


apollo/integrations/db/salesforce_data_cloud_proxy_client.py
```diff
@@ -396,6 +422,142 @@ def list_tables(self, dataspace: str | None = None) -> list[dict]:
...
+
+        Raises ``RuntimeError`` with the HTTP status in the ``code NNN`` format
+        on any non-200 response. Both HTTP calls flow through a single
+        ``_CapturingSession`` so the redacted response body is surfaced in the
+        exception message and structured log record on failure.
+        """
+        domain = self._credentials.domain
+        session = _CapturingSession()
+
+        logger.info(
```
genehynson: might be noisy - debug instead?

genehynson (approved):

apollo/integrations/db/salesforce_data_cloud_proxy_client.py
walles2: Fixed in 0eb575f — demoted to `logger.debug`. The success-path info log already records the outcome (count + names), and error paths still emit at WARNING, so the start-of-call line was redundant in steady state.

PR #300, opened 2026-05-20
feat(credentials): cache externally-fetched secrets across operations

mrostan:
## Summary

Cache the resolved external credentials (ASM / AKV / GSM / etc.) in a process-wide TTL cache so we make at most one network round-trip per credential identifier per TTL window, instead of one per operation.

## Motivation

While investigating a customer agent we observed Databricks ops taking 15–24s end-to-end, with the time spent entirely between credentials fetch start and completion. The same agent's Tableau ops, backed by file-based credentials, were sub-second. ASM was being hit on every op (fresh `boto3.Session` + fresh `secretsmanager` client per call).

## What changes

- New module `apollo/credentials/external_credentials_cache.py`
  - `TTLCache(maxsize=128, ttl=300s)` keyed by sha256 of the credentials dict with `connect_args` removed — request-specific values like warehouse id / http_path don't change which secret to fetch
  - Returns a deep copy on every get so the in-place mutation in `_merge_connect_args` doesn't corrupt the cached value
  - TTL configurable via `MCD_EXTERNAL_CREDENTIALS_CACHE_TTL_SECONDS` (`0` disables caching but keeps the diagnostic log). Value is read once at module import; restart required to change.
  - Documented caveat: `cachetools.@cached` releases the lock between lookup and loader, so concurrent first-misses for the same key all invoke the loader. In practice the egress agent's pattern is "one first call, then concurrent warm hits", so the herd is small and the correctness cost is zero (last write wins).
- `BaseCredentialsService`
  - Routes through `_load_external_credentials_cached` in `get_credentials`
  - `__init__(provider_name: str)` — required, no default. Forgetting the arg in a new subclass fails loudly at instantiation.
  - Passthrough `BaseCredentialsService` bypasses the cache (no network I/O, no benefit, would pin request-specific dicts)
- All five providers (`asm`, `akv`, `gsm`, `env_var`, `file`) declare their `provider_name` via `super().__init__(...)`
- Each `load_cached()` call emits one INFO log so we can verify the cache in production and triage residual slowness:

  ```
  Loaded external credentials, provider=aws_secrets_manager, cache=hit, duration_s=0.000
  Loaded external credentials, provider=aws_secrets_manager, cache=miss, duration_s=0.412
  Loaded external credentials, provider=aws_secrets_manager, cache=disabled, duration_s=0.411
  Loaded external credentials, provider=aws_secrets_manager, cache=miss-failed, duration_s=15.012
  ```

## Test plan

- [x] 24 new unit tests covering cache key stability/separation, hit/miss, failure not cached, deep-copy semantics, log shape, TTL=0 disable
- [x] All existing credentials tests still pass (split a multi-scenario env_var test that used the same key for 3 different env values in one method)
- [x] `pytest tests/` — **1057 passed**
- [x] `black --check`, `pyright` — clean
- [x] Smoke tested on dev EC2 agent (`977f5159-4bcd-4370-9e59-2f3d195d0afc`): 1 `cache=miss` (duration_s=0.172) followed by 27 `cache=hit` (duration_s=0.000) on subsequent ops, 96.4% hit rate — `Found credentials from IAM Role` events dropped from ~130 in 2 min on the old image to 1 in the same window on the new image

### DIFF
--- a/apollo/credentials/akv.py
+++ b/apollo/credentials/akv.py
@@ -14,6 +14,9 @@ class AzureKeyVaultCredentialsService(BaseCredentialsService):
     Credentials service that fetches credentials from a secret in Azure Key Vault.
     """
 
+    def __init__(self):
+        super().__init__(provider_name="azure_key_vault")
+
     def _load_external_credentials(self, credentials: dict) -> dict:
         secret_name = credentials.get(SECRET_NAME)
         if not secret_name:
--- a/apollo/credentials/asm.py
+++ b/apollo/credentials/asm.py
@@ -11,6 +11,9 @@ class AwsSecretsManagerCredentialsService(BaseCredentialsService):
     Credentials service that fetches credentials from AWS Secrets Manager.
     """
 
+    def __init__(self):
+        super().__init__(provider_name="aws_secrets_manager")
+
     def _load_external_credentials(self, credentials: dict) -> dict:
         secret_name = credentials.get(SECRET_NAME)
         if not secret_name:
--- a/apollo/credentials/base.py
+++ b/apollo/credentials/base.py
@@ -1,14 +1,42 @@
+from apollo.credentials.external_credentials_cache import load_cached
+
+
 class BaseCredentialsService:
     """
     Base class for credentials services, provides default behavior of
     expecting warehouse credentials to be included in the request.
+
+    Subclasses pass a short ``provider_name`` identifier to ``__init__``
+    (e.g. ``aws_secrets_manager``, ``azure_key_vault``). The name appears in
+    the cache log lines emitted by :func:`load_cached` and is purely for
+    observability — it doesn't affect routing. Making it a required
+    constructor argument means a new subclass that forgets to supply it
+    fails loudly at instantiation rather than silently logging a generic
+    label.
     """
 
+    def __init__(self, provider_name: str):
+        self.provider_name = provider_name
+
     def get_credentials(self, credentials: dict) -> dict:
-        external_credentials = self._load_external_credentials(credentials)
+        external_credentials = self._load_external_credentials_cached(credentials)
         return self._merge_connect_args(
             incoming_credentials=credentials,
             external_credentials=external_credentials,
+        )
+
+    def _load_external_credentials_cached(self, credentials: dict) -> dict:
+        """Cache the result of ``_load_external_credentials`` across operations.
+
+        The default ``BaseCredentialsService`` implementation is a passthrough
+        that does no network I/O — caching it adds no value and would pin
+        request-specific dicts in memory, so we short-circuit it. Subclasses
+        that talk to ASM / AKV / GSM benefit from the cache.
+        """
+        if type(self) is BaseCredentialsService:
+            return self._load_external_credentials(credentials)
+        return load_cached(
+            credentials, self._load_external_credentials, self.provider_name
         )
 
     def _load_external_credentials(self, credentials: dict) -> dict:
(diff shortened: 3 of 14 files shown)


mrostan:
## Review changes summary

Two rounds of independent review run. All non-deferred items are addressed.

### Round 1 — addressed in 600fb5a (docs only)

- [x] **TTL is read at module import.** Documented in the cache module docstring so on-call doesn't expect a live env var flip to take effect; restart is required.
- [x] **`_CACHE_MAX_SIZE = 128` rationale.** Comment added explaining sizing (one entry per (provider, secret name, region, assumable role) tuple; well above the working set for any realistic agent).
- [x] **`type(self) is BaseCredentialsService` gate.** Comment added explaining why this is the right gate (not `isinstance`) — it bypasses the cache only for direct uses of the passthrough base.
- [x] **PR description.** Dropped the forward-reference to the OAuth-cache PR style (that one isn't being merged).

### Round 1 — accepted as is (rationale in module docstring)

- **Cachetools thundering herd**: `@cached` releases the lock between lookup and loader, so N concurrent first-misses for the same key all invoke the loader. In practice this egress agent's pattern is "one first call, then concurrent warm hits", so the herd is small and correctness is unaffected (all loaders return the same value). Per-key locking would add complexity not justified by observed load.
- **env_var / file providers cached for 300s**: consistent with ASM/AKV/GSM. A rotation propagation lag of up to 300s is the same trade-off across providers.

### Round 2 — addressed in 4538afe

- [x] **`try/finally` for the miss-failed log gap.** Previously, exceptions raised before the loader was invoked (e.g. inside the cache key hash, lock acquisition, or `deepcopy` on success) went completely unlogged. Now every path emits a log; introduces a fifth state `cache=error` for "machinery raised" vs. `cache=miss-failed` for "loader raised". Both states are pinned by tests.
- [x] **Stale docstring on `_load_and_cache`**: it claimed the cache key uses the loader, but the key function (by design) ignores it. Clarified.
- [x] **Explicit nested-mutation deep-copy test.** The existing test only mutated a top-level key, which a shallow copy would also pass. Added `test_nested_mutation_does_not_corrupt_cached_value` that fails under shallow copy.
- [x] **`provider_name` → `_provider_name`.** Constructor kwarg stays public; instance attribute renamed to signal it's an implementation detail of the cache logging path, not part of the service's contract.

### Round 2 — deferred

- **Structured `extra={...}` logging.** Nice-to-have so Datadog can promote `provider`, `cache`, `duration_s` to first-class JSON attributes instead of requiring regex facet extraction. Deferred: the egress agent (hermes) doesn't propagate `extra` to its log shipper yet. Will add `extra` support in hermes in a follow-up PR, then convert the cache logs.

### Round 2 — accepted as is

- **`conftest.py` autouse scope (`tests/credentials/` only)** — fine as long as cache-using tests live under that directory; documented constraint.
- **`KeyboardInterrupt` / `SystemExit` not exercised** — `except Exception` correctly doesn't catch them and the cache stays clean; no test pinning. Standard Python behavior, not worth a regression test.
- **`_merge_connect_args` aliasing of incoming `connect_args`** — pre-existing behavior; the deep-copy boundary only isolates the *external* half. Documented in the PR description.

### Verification

- Tests: **69 in `tests/credentials/`**, full suite still passes
- `black --check`, `pyright` — clean
- Smoke tested on dev EC2 agent `977f5159-…`: 1 `cache=miss` (0.172s) → 27 `cache=hit` (0.000s), `Found credentials from IAM Role` events dropped from ~130 in 2 min on the old image to 1

walles2 (approved):

97nitt (approved):

PR #308, opened 2026-05-28
feat(security): block cloud metadata services and other non-routable IPs

mrostan:
## Summary

Blocks the agent's outbound HTTP integrations from connecting to cloud
metadata services (169.254.169.254 and friends) and other non-routable
addresses. Flagged by customer security teams as an SSRF risk in the
generic ``http`` connection type.

## Policy

The guard runs inside a wrapper around ``urllib3.util.connection.create_connection``.
The wrapper is a no-op (passthrough) when no policy is active on the current
thread, so unrelated urllib3 consumers in the same process (Snowflake connector,
GCS/Azure SDKs, OpenTelemetry exporter, etc.) are unaffected. Two policy tiers
share one implementation:

| Tier | Used by | Blocks |
|---|---|---|
| **Default** | ``HttpProxyClient.do_request``, ``TableauProxyClient.api_request`` | `169.254.0.0/16` (IMDS), `fe80::/10` (IPv6 link-local), `127.0.0.0/8` + `::1/128` (loopback). **RFC1918 allowed** for VPC / Private Link traffic. |
| **Strict** | ``HttpProxyClient.download_*`` (pre-signed URLs from further upstream) | All non-public IPs (RFC1918, link-local, loopback, multicast, reserved, unspecified) + non-HTTPS. Replaces ``_assert_safe_download_url``. |
| **Network troubleshooting validators** | ``ValidateNetwork.validate_http_connection`` (via ``safe_request``); ``validate_tcp_open_connection`` and ``validate_telnet_connection`` (via ``assert_safe_destination``) | Same default-tier rules. The troubleshooting endpoint can no longer be used to reach cloud metadata services even from a Monte Carlo operator session. |

Operators can extend the default block list via the
``MCD_HTTP_BLOCKED_CIDRS`` env var (e.g.
``MCD_HTTP_BLOCKED_CIDRS="100.64.0.0/10,10.50.0.0/16"``). Invalid entries are
logged and skipped at import time. The extension applies under both tiers.

Operators can also enforce HTTPS in the default tier via
``MCD_HTTP_REQUIRE_HTTPS=true`` (default ``false`` — preserves current behavior
for existing customer integrations that use plain HTTP over private networks).

## Design notes

- **Redirects:** every hop goes through ``create_connection``, so the guard
  runs on each redirect target without any ``requests.Session`` subclassing.
- **Multi-IP fallback:** when a hostname resolves to multiple addresses, the
  hook iterates them, validating each just before its connect attempt and
  falling back on transient connect failure — matches urllib3's native
  behavior.
- **Exception propagation:** ``HttpClientError`` raised from inside
  ``create_connection`` propagates cleanly through urllib3 and ``HTTPAdapter``
  (neither catches arbitrary subclasses), so callers see the original
  exception type rather than a wrapped ``ConnectionError``.
- **Composability:** ``safety_policy`` saves/restores prior thread-local
  state on entry/exit, so nesting contexts (or test re-use of the same
  thread) doesn't silently deactivate an outer policy.
- **Idempotency:** the urllib3 hook install carries a marker attribute and
  no-ops on re-import, preventing wrapper stacking.
- **Threading:** the policy is ``threading.local`` — fan-out to a worker
  thread (``ThreadPoolExecutor``, ``run_in_executor``) requires re-entering
  ``safety_policy`` on the worker.

## Out of scope

- **DB clients** (Snowflake, BigQuery, Redshift, Starburst, Salesforce Data
  Cloud, etc.) speak DB-specific wire protocols; a metadata-service IP fails
  handshake rather than serving credentials. Different threat model.
- **CTP transforms** (``oauth``, ``resolve_informatica_session``) don't return
  response bodies to the caller, so SSRF doesn't yield meaningful exfiltration
  here. The real risk in these is generic DC-supplied URL substitution for
  credential leak, which this guard doesn't address (and which RFC1918-allow
  in the default tier wouldn't address either).

## Files changed

- New: ``apollo/integrations/http/url_safety.py`` — the helper (
  ``safety_policy`` context manager, ``safe_request`` wrapper, ``HttpClientError``,
  ``_safe_create_connection`` hook).
- New: ``apollo/integrations/http/CLAUDE.md`` — architecture, call-site
  guidance, threading limitation, out-of-scope notes.
- ``apollo/integrations/http/http_proxy_client.py`` — ``do_request`` uses
  ``safe_request``; ``_open_download_response`` uses
  ``safety_policy(url, strict_ip_policy=True, https_only=True)``;
  ``_assert_safe_download_url`` deleted; explicit
  ``HttpClientError as HttpClientError`` re-export.
- ``apollo/integrations/tableau/tableau_proxy_client.py`` — ``api_request``
  uses ``safe_request``.
- ``README.md`` + ``.envrc.example`` — document
  ``MCD_HTTP_BLOCKED_CIDRS`` and ``MCD_HTTP_REQUIRE_HTTPS``.
- New: ``tests/test_url_safety.py`` — hook install, passthrough,
  default + strict tiers, multi-IP fallback, env-var parsing, context
  manager, nested-context save/restore, cross-thread isolation,
  ``MCD_HTTP_REQUIRE_HTTPS`` behavior.
- ``tests/test_http_client.py`` — download-tier tests updated for new
  error wording; ``TestHttpClientSsrfGuard`` for the
  ``do_request`` integration path.
- ``tests/test_tableau_client.py`` — assertion updated to match
  ``requests.request`` positional call shape (now goes through
  ``safe_request``).
- ``apollo/validators/validate_network.py`` — ``validate_http_connection``
  uses ``safe_request``; ``validate_tcp_open_connection`` and
  ``validate_telnet_connection`` call ``assert_safe_destination`` before
  opening the socket / Telnet session. ``perform_dns_lookup`` is unchanged
  (reconnaissance-only, no connection).
- ``tests/test_health_network.py`` — existing tests use a public IP literal
  so the new pre-flight short-circuits; new SSRF regression tests for the
  three guarded validators.
- ``apollo/integrations/http/url_safety.py`` — added the
  ``assert_safe_destination(host, port, *, strict_ip_policy=False)`` helper.

## Test plan

- [x] All affected unit tests pass locally (1173 passed in the full suite;
  the 3 pre-existing collection errors are unrelated — missing ODBC driver
  on the dev machine).
- [x] Ran `/code-review` (two rounds + a third round for the
  ``validate_network`` follow-up)
- [ ] CI green.
- [ ] Manual verification: confirm a sample integration (Tableau, MuleSoft)
  still works against a real endpoint.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,6 +8,6 @@ repos:
       types_or: [python, pyi]
 
   - repo: https://github.com/RobertCraigie/pyright-python
-    rev: v1.1.337
+    rev: v1.1.407
     hooks:
     - id: pyright
--- a/apollo/integrations/http/url_safety.py
+++ b/apollo/integrations/http/url_safety.py
@@ -248,6 +248,17 @@ def assert_safe_destination(
         the block list applies uniformly — but the policy structure
         (e.g. adding a new early-exit check) must be kept in sync
         manually.
+
+        Note the two intentionally differ on mixed record sets: this
+        function is strict and rejects if *any* resolved address is
+        blocked, because it only returns a validated IP and does not
+        connect. ``_safe_create_connection`` validates lazily, in
+        ``getaddrinfo`` order, only the addresses it actually attempts —
+        a blocked address that appears *after* one it connects to
+        successfully is never reached and never checked. That is
+        deliberate (it mirrors urllib3's own fall-through-on-failure
+        behavior); don't "fix" the divergence by forcing either side to
+        match the other.
     """
     if not host:
         raise HttpClientError("destination refuses '<empty>' host")
--- a/apollo/integrations/http/url_safety.py
+++ b/apollo/integrations/http/url_safety.py
@@ -35,7 +35,14 @@ Operators can extend the default block list via the
 ``MCD_HTTP_BLOCKED_CIDRS`` env var: a comma-separated list of CIDRs
 (e.g. ``"100.64.0.0/10,10.50.0.0/16"``). Invalid entries are logged
 and skipped at import time so a typo never crashes the agent. The
-env-var extension applies on top of both policy tiers.
+env-var extension applies on top of both policy tiers. Read once at
+module-import time — changes require a process restart.
+
+Operators can also enforce HTTPS in the default policy tier by setting
+``MCD_HTTP_REQUIRE_HTTPS=true``. The strict download tier (which passes
+``https_only=True`` explicitly) is unaffected — it already requires
+HTTPS unconditionally. Only the default tier picks up the env-var policy.
+Read once at module-import time — changes require a process restart.
 
 The ``create_connection`` wrapper is **always installed** at import
 time but is a passthrough whenever no policy is active on the current
@@ -86,6 +93,7 @@ class HttpClientError(Exception):
 
 
 _ENV_EXTRA_BLOCKED_CIDRS = "MCD_HTTP_BLOCKED_CIDRS"
+_ENV_REQUIRE_HTTPS = "MCD_HTTP_REQUIRE_HTTPS"
 
 _DEFAULT_BLOCKED_CIDRS: Tuple[str, ...] = (
     "169.254.0.0/16",
@@ -123,10 +131,39 @@ def _load_extra_blocked_networks() -> List[_Network]:
     return _parse_cidrs(tuple(raw.split(",")), source=_ENV_EXTRA_BLOCKED_CIDRS)
 
 
+def _load_bool_env(name: str, *, default: bool) -> bool:
+    raw = os.environ.get(name, "").strip().lower()
+    if raw in ("true", "1", "yes", "on"):
+        return True
+    if raw in ("false", "0", "no", "off", ""):
+        return default if raw == "" else False
+    _logger.warning(
+        "url_safety: ignoring invalid value for %s: '%s' — using default %s",
+        name,
+        raw,
+        default,
+    )
+    return default
+
+
 _DEFAULT_NETWORKS: List[_Network] = _parse_cidrs(
     _DEFAULT_BLOCKED_CIDRS, source="defaults"
 )
 _EXTRA_NETWORKS: List[_Network] = _load_extra_blocked_networks()
+_REQUIRE_HTTPS_BY_DEFAULT: bool = _load_bool_env(_ENV_REQUIRE_HTTPS, default=False)
+
+
+def _log_startup_summary() -> None:
+    _logger.info(
+        "url_safety: SSRF guard initialized — %d default CIDR(s), %d extra CIDR(s) from %s, https_only=%s",
+        len(_DEFAULT_NETWORKS),
+        len(_EXTRA_NETWORKS),
+        _ENV_EXTRA_BLOCKED_CIDRS,
+        _REQUIRE_HTTPS_BY_DEFAULT,
+    )
+
+
+_log_startup_summary()
 
 
 def _ip_is_rejected(ip: _Address, *, strict_ip_policy: bool) -> bool:
@@ -159,7 +196,8 @@ def _assert_safe_url_scheme_and_host(url: str, *, https_only: bool) -> None:
     block) but produces a clearer error message at the URL layer.
     """
     parts = urlsplit(url)
-    allowed_schemes = ("https",) if https_only else ("http", "https")
+    effective_https_only = https_only or _REQUIRE_HTTPS_BY_DEFAULT
+    allowed_schemes = ("https",) if effective_https_only else ("http", "https")
     if parts.scheme not in allowed_schemes:
         raise HttpClientError(f"request refuses unsupported scheme '{parts.scheme}'")
     host = (parts.hostname or "").lower()
@@ -187,6 +225,13 @@ def _assert_safe_url_scheme_and_host(url: str, *, https_only: bool) -> None:
 # ``requests.Session`` subclassing.
 
 _policy = threading.local()
+# We hook urllib3.util.connection.create_connection because it is the
+# function every HTTP(S) connection ultimately calls for the TCP socket
+# create+connect step. This is internal urllib3 API; the hook has been
+# verified against urllib3 1.26.x and 2.x. If a future urllib3 version
+# reorganizes connection creation away from this function, the hook
+# will silently become a no-op — `test_hook_is_installed` is the
+# canary for that.
 _original_create_connection = _urllib3_connection.create_connection
 
 
@@ -206,6 +251,11 @@ def _safe_create_connection(address: Tuple[str, int], *args: Any, **kwargs: Any)
     if literal is not None:
         if _ip_is_rejected(literal, strict_ip_policy=strict):
             raise HttpClientError(f"request refuses blocked address: {literal}")
+        # Pass the original `address` tuple through unchanged: when the IP is
+        # not rejected, urllib3 sees exactly what it would have without the
+        # wrapper. Reconstructing here (e.g. `(str(literal), port)`) would
+        # work — but minimizing interference avoids subtle interpretation
+        # drift between the validated and pass-through paths.
         return _original_create_connection(address, *args, **kwargs)
 
     # Hostname: resolve, then validate-and-connect in a single pass.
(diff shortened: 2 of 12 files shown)


mrostan:
<details>
<summary>Code Review — 15 findings (15 fixed, 0 skipped)</summary>

**Reviewed:** 2026-05-28 UTC | **Reviewers:** security, devdocs, correctness, testing, architecture | **Scope:** full PR

| F1 | BLOCKER | `safety_policy` does not save/restore prior thread-local state — nesting silently deactivates outer policy | ✅ Fixed — snapshot and restore prior `active`/`strict_ip_policy`/`https_only` in `finally` |
| F2 | ISSUE | `_policy.https_only` stored but never consulted at TCP layer — dead state or missing enforcement | ✅ Fixed — removed unread field; docstring now states `https_only` enforced only at URL pre-flight; strict tier must use `allow_redirects=False` |
| F3 | ISSUE | Redirect-following guard advertised but never exercised by a test | ✅ Fixed — added integration test following redirect through the wrapper |
| F4 | ISSUE | `do_request` SSRF integration not directly tested end-to-end | ✅ Fixed — added `do_request` test targeting a default-blocked IP through real urllib3 |
| F5 | ISSUE | urllib3 monkey-patch is not idempotent and has no opt-out | ✅ Fixed — idempotency guard via `_mc_ssrf_guard` sentinel attribute |
| F6 | ISSUE | `threading.local` policy silently bypassed under thread/async executors | ✅ Fixed — documented "must be entered on issuing thread" constraint in `safety_policy` docstring |
| F7 | ISSUE | `MCD_HTTP_BLOCKED_CIDRS` silently degrades on parse errors | ✅ Fixed — startup summary log enumerates loaded and skipped entries |
| F8 | SUGGESTION | Missing `apollo/integrations/http/CLAUDE.md` | ✅ Fixed — added CLAUDE.md covering hook design, threading.local contract, policy tiers, and env vars |
| F9 | SUGGESTION | `MCD_HTTP_BLOCKED_CIDRS` not documented in README / `.envrc.example` | ✅ Fixed — added env var entry to README |
| F10 | SUGGESTION | Thread-safety / `threading.local` isolation not tested | ✅ Fixed — added concurrent-thread policy isolation test |
| F11 | SUGGESTION | urllib3 version coupling not documented | ✅ Fixed — added version comment and module-load assertion |
| F12 | SUGGESTION | `_EXTRA_NETWORKS` evaluated once at import with no startup summary | ✅ Fixed — info-level log at module load; docstring updated |
| F13 | SUGGESTION | `HttpClientError` re-export from `http_proxy_client.py` is implicit | ✅ Fixed — `from .url_safety import HttpClientError as HttpClientError` (PEP 484 re-export) |
| F14 | SUGGESTION | HTTP scheme allowed in default tier with no operator opt-in for HTTPS enforcement | ✅ Fixed — added `MCD_HTTP_REQUIRE_HTTPS` env var (default `false`); documented alongside `MCD_HTTP_BLOCKED_CIDRS` |
| F15 | NIT | IP-literal branch passes original `address` tuple rather than normalized literal | ✅ Fixed — added comment explaining intentional pass-through to minimize urllib3 interference |

👉 [Fix commit](https://github.com/monte-carlo-data/apollo-agent/commit/470b412)

</details>

mrostan:
<details>
<summary>Code Review — 11 findings (11 fixed, 0 skipped)</summary>

**Reviewed:** 2026-05-28 UTC | **Reviewers:** security x2, devdocs x2, correctness x2, testing x2, architecture x2 | **Scope:** incremental from `108d9c6`

| F1 | ISSUE | CLAUDE.md tier-comparison table missing `MCD_HTTP_REQUIRE_HTTPS` in default-tier row | ✅ Verified |
| F2 | ISSUE | `_load_bool_env` parser has no direct tests — integration tests bypass it via direct patch | ✅ Verified |
| F3 | ISSUE | `_mc_ssrf_guard` marker uses a generic `True` — spoofable by attribute-name collision | ✅ Verified |
| F4 | ISSUE | Idempotency comment inaccurate under `importlib.reload()` | ✅ Verified |
| F5 | SUGGESTION | tearDown only resets `_policy.active`, not `_policy.strict_ip_policy` | ✅ Verified |
| F6 | SUGGESTION | No regression test for nested `safety_policy` where inner context raises | ✅ Verified |
| F7 | SUGGESTION | Threaded tests patch a module-level attribute from inside worker threads | ✅ Verified |
| F8 | SUGGESTION | CLAUDE.md doesn't note that `HttpProxyClient` subclasses inherit the SSRF guard automatically | ✅ Verified |
| F9 | SUGGESTION | Pre-flight check runs before snapshot — fragile invariant under future refactor | ✅ Verified |
| F10 | NIT | `_load_bool_env` empty-string branch is convoluted | ✅ Verified |
| F11 | NIT | `test_do_request_with_rfc1918_url_is_allowed` name doesn't specify which tier | ✅ Verified |

👉 [Fix commit](https://github.com/monte-carlo-data/apollo-agent/commit/90a8114)

</details>

.pre-commit-config.yaml
```diff
@@ -8,6 +8,6 @@ repos:
       types_or: [python, pyi]
 
   - repo: https://github.com/RobertCraigie/pyright-python
-    rev: v1.1.337
+    rev: v1.1.407
```
mrostan, line 11: use the same version used in CI

mrostan:
<details>
<summary>Code Review — 9 findings (9 fixed, 0 skipped)</summary>

**Reviewed:** 2026-05-29 UTC | **Reviewers:** security x2, devdocs x2, correctness x2, testing x2, architecture x2 | **Scope:** incremental from `470b412`

| F1 | BLOCKER | DNS-rebinding TOCTOU between `assert_safe_destination` and `socket.connect_ex` / `Telnet` | ✅ Resolved once in helper, callers connect by validated IP |
| F2 | ISSUE | HTTP path missing `HttpClientError → ConnectionFailedError` wrapper | ✅ Wrapper added to `_internal_validate_http_connection` |
| F3 | ISSUE | AWS IMDSv2 IPv6 address `fd00:ec2::254` not in default block list | ✅ Added `fd00:ec2::/64` to `_DEFAULT_BLOCKED_CIDRS` |
| F4 | SUGGESTION | Missing regression test: `validate_telnet_connection` rejects `localhost` | ✅ Test added mirroring TCP sibling |
| F5 | SUGGESTION | Module/class docstrings and cross-reference comment behind round-3 changes | ✅ Docstrings and cross-reference comments updated |
| F6 | NIT | `localhost.` (trailing dot) bypasses string fast-path | ✅ Trailing dot stripped before comparison |
| F7 | NIT | IPv6 zone IDs make fast-path comment slightly inaccurate | ✅ Zone ID stripped before `ipaddress.ip_address` |
| F8 | NIT | `assertIsNone(assert_safe_destination(...))` is a weak success assertion | ✅ Assertion style updated |
| F9 | NIT | Missing test for bracketed IPv6 literal `[::1]` | ✅ Test added |

👉 [Fix commit](https://github.com/monte-carlo-data/apollo-agent/commit/16f697a)

</details>

apollo/integrations/http/url_safety.py
```diff
@@ -0,0 +1,509 @@
...
+        HttpClientError: empty/localhost host, blocked IP literal, blocked
+            IP in the resolved record set, or DNS resolution failure.
+
+    See also:
+        ``_safe_create_connection`` — the connection-time sibling that
+        applies the same block list inside urllib3's TCP-connect path.
+        These two functions share ``_ip_is_rejected``, so any change to
+        the block list applies uniformly — but the policy structure
+        (e.g. adding a new early-exit check) must be kept in sync
+        manually.
```
sassmith, line 250: Nit: the docstring says the policy structure "must be kept in sync manually" with _safe_create_connection, but they handle mixed records differently. `assert_safe_destination` rejects if any resolved IP is blocked, while `_safe_create_connection` validates lazily and skips blocked IPs it never tries to connect to. Might be worth updating the comment to call out the intentional difference so a future reader doesn't try to "fix" the divergence?

sassmith (approved):

apollo/integrations/http/url_safety.py
mrostan: Good catch — done in `605e184` (pushed). Extended the `See also:` note to spell out the intentional divergence: `assert_safe_destination` is strict (rejects if *any* resolved address is blocked, since it only validates and never connects), while `_safe_create_connection` validates lazily in `getaddrinfo` order and never checks addresses past a successful connect — mirroring urllib3's fall-through-on-failure behavior. The note now explicitly says not to reconcile the two.

PR #310, opened 2026-05-28
Path-based session-header routing for v2 + v3 Informatica endpoints

97nitt:
## Changes

Enables the agent-side `InformaticaProxyClient` to drive Informatica V3 API endpoints (e.g. `/public/core/v3/objects`) in addition to the existing V2 endpoints (`/api/v2/...`), without an architectural split or per-call routing decision from the DC.

Each Informatica API surface reads the session token from a different header:

- V2 endpoints (`/api/v2/...`) read `icSessionId`.
- V3 endpoints (`/public/core/v3/...`) read `INFA-SESSION-ID`.

`do_http_request` now picks the right header based on the request path so only the header that endpoint actually reads goes on the wire. The session token continues to be sourced from `connect_args.session_id` (resolved upstream by the `resolve_informatica_session` CTP transform).

This is the agent-side prerequisite for the Informatica v2 path-info enrichment cross-repo work — the DC's path enrichment fetches `/public/core/v3/objects?q=type=='MTT'` for self-hosted Informatica customers. Sibling PRs:

- `data-collector` / services/orchestrator: [#2369](https://github.com/monte-carlo-data/data-collector/pull/2369) — orchestrator data-cache endpoints (in review)
- `monolith-django` #13691 — loader + catalog metadata + search read surface (in dev for regression check)
- `data-collector`: [#2370](https://github.com/monte-carlo-data/data-collector/pull/2370) — DC-side path enrichment + `list_objects` agent-proxy method (depends on this)

## Design notes

- **Path-based routing in the proxy client, not per-version proxy clients.** A single Informatica connection hits both API surfaces in the same collection cycle (V2 for MTT detail / activity logs, V3 for the path enrichment), so splitting at the connection-type level would force two proxy instances per connection without removing the routing decision.
- **No per-method wrapper.** The agent's HTTP proxy is intentionally generic (`do_http_request(path, ...)`); typed wrappers live on the DC-side `InformaticaAgentProxyClient`. The DC's `list_objects` method calls `do_http_request(path="/public/core/v3/objects", params=...)` and the agent automatically sends the correct header.
- **No header auto-attach in HttpProxyClient.** Since the session header is path-dependent, the proxy no longer relies on `HttpProxyClient`'s `auth_header` credential — it injects the header explicitly per call via `additional_headers`. Caller-supplied `additional_headers` continue to win on collision so deliberate overrides aren't silently masked.

## Testing

- `pytest tests/test_informatica_proxy_client.py` — 13/13 passing.
- `pytest tests/ctp/test_informatica_ctp.py tests/ctp/test_informatica_v2_ctp.py tests/ctp/test_resolve_informatica_session.py` — 46/46 passing (regression check on the upstream session-resolve transforms).
- `black --check` and `pyright` — clean.

New / updated tests:

- `test_v2_path_sends_only_ic_session_id` — V2 paths get `icSessionId` and *not* `INFA-SESSION-ID` on the wire.
- `test_v3_path_sends_only_infa_session_id` — V3 paths get `INFA-SESSION-ID` and *not* `icSessionId`. Locks in the path-based routing so a future refactor doesn't silently regress to dual-header or wrong-header behavior.
- `test_caller_headers_win_on_session_header_collision` — caller-supplied `additional_headers` override the proxy's session-header pick.
- `test_v2_connection_type_uses_same_proxy_client` extended to assert the v2 connection type also sends only `icSessionId`.

## Verification after deploy

The target customer's agent needs to be running this version before the DC's path-enrichment PR ships. Coordinate the agent upgrade per the cross-repo plan.

1. After agent upgrade, hit a V2 endpoint through the agent (e.g., MTT list) — expect 200 with `icSessionId` on the wire (no regression).
2. Hit a V3 endpoint through the agent (`/public/core/v3/objects?q=type=='MTT'`) — expect 200 with `INFA-SESSION-ID` on the wire. End-to-end exercise lands with the DC path-enrichment PR.

### DIFF
--- a/apollo/integrations/http/informatica_proxy_client.py
+++ b/apollo/integrations/http/informatica_proxy_client.py
@@ -86,6 +86,10 @@ class InformaticaProxyClient(BaseProxyClient):
         """
         Execute an HTTP request against the Informatica API base URL.
 
+        The session header is selected by path prefix: ``/public/core/v3/...``
+        gets ``INFA-SESSION-ID``; everything else gets ``icSessionId``.
+        Caller-supplied ``additional_headers`` override on collision.
+
         :param path: Path to append to the API base URL (e.g., "/v2/jobs"). Must start with "/".
         :param http_method: HTTP method (GET, POST, PUT, DELETE, etc.)
         :param payload: Optional JSON payload for the request body
--- a/apollo/integrations/http/informatica_proxy_client.py
+++ b/apollo/integrations/http/informatica_proxy_client.py
@@ -14,16 +14,22 @@ class InformaticaProxyClient(BaseProxyClient):
     resolve_informatica_session transform) to contain a pre-resolved session:
 
         connect_args:
-            session_id    (required): icSessionId from the Informatica login response
+            session_id    (required): session token from the Informatica login response
             api_base_url  (required): API base URL from the Informatica login response
 
     All authentication — V2 password, V3 password, or JWT loginOAuth — is handled
     upstream in the CTP pipeline. This client is auth-method agnostic.
 
-    The icSessionId header is always used for API calls regardless of how the session
-    was obtained. This is intentional: the DC currently calls only V2 API endpoints,
-    which require icSessionId (not INFA-SESSION-ID).
+    The Informatica session header is selected per request based on the path:
+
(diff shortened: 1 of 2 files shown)


mrostan (approved):
lgtm

PR #319, opened 2026-06-05
fix(ctp): close concurrent-init race in CTP registries (YET-1420)

walles2:
## Summary

Closes the concurrent-init race in `TransformRegistry` and `CtpRegistry` that produced the
`Unknown transform type: 'resolve_databricks_oauth'` errors on Haleon's Azure agent (and
Ensemble HP, Wesco, Mercedes Benz — all Azure agents under multi-threaded WSGI).

[YET-1420](https://linear.app/montecarloai/issue/YET-1420)

## Root cause

`_ensure_initialized()` flipped `_initialized = True` **before** running `_discover()`:

```python
def _ensure_initialized() -> None:
    global _initialized
    if not _initialized:
        _initialized = True   # ← flag set first
        _discover()           # ← imports happen here
```

Under Azure WSGI threading, a second thread arriving during the cold-start window would see
`_initialized == True` while the first thread's `_discover()` was still running, then look up
a transform that hadn't been registered yet. Same pattern existed in both `CtpRegistry` and
`TransformRegistry`.

## Why we know it's a race (not a deterministic import-time failure)

Slack thread reported two **different** partial-registry states in the error message: 3 of 15
and 9 of 15. A deterministic import-time failure (e.g. `databricks-sdk` mismatch crashing
`resolve_databricks_oauth.py` at import) would always halt at the same module — it could not
produce two different prefixes of the import order. Different prefixes mean different timing.

The "9 of 15" case in particular maps exactly to the import list: `resolve_msal_token` is the
9th import; `resolve_databricks_oauth` (the symptom) is the 10th. Thread B arrived in that gap.

## Fix

Double-checked locking. Post-init path stays lock-free; cold-start window is guarded so the
second thread either waits or also runs `_discover()` — never observes a partial registry.

```python
def _ensure_initialized() -> None:
    global _initialized
    if _initialized:
        return
    with _init_lock:
        if not _initialized:
            _discover()
            _initialized = True
```

## Test plan

- [x] `tests/ctp/test_registry_concurrent_init.py` drives the race deterministically — Thread A's
      `_discover()` is wrapped to block partway via an Event, Thread B's lookup races in, assertion
      catches partial-registry observations. Reproduces the exact Haleon error against the prior
      code and passes against the fix.
- [x] Full CTP suite: **505 passed**, 0 failed (including the 2 new tests).
- [x] `black --check` clean on changed files.
- [x] `pyright` clean (0 errors, 0 warnings).
- [ ] Verify on Azure dev environment: cold-start a fresh container and hit it with concurrent
      requests; confirm the `Unknown transform type` error no longer occurs.

## Files

| File | Change |
|---|---|
| `apollo/integrations/ctp/transforms/registry.py` | DCL fix on `_ensure_initialized()` |
| `apollo/integrations/ctp/registry.py` | Identical DCL fix on the connector registry |
| `tests/ctp/test_registry_concurrent_init.py` | **NEW** — parametrized concurrent-init regression suite |

### DIFF
--- a/apollo/integrations/ctp/registry.py
+++ b/apollo/integrations/ctp/registry.py
@@ -1,4 +1,5 @@
 # apollo/integrations/ctp/registry.py
+import threading
 from typing import Any
 
 from apollo.integrations.ctp.errors import CtpPipelineError
@@ -7,6 +8,7 @@ from apollo.integrations.ctp.pipeline import CtpPipeline
 
 _ATTR_CONNECT_ARGS = "connect_args"
 _initialized: bool = False
+_init_lock = threading.Lock()
 
 
 def _discover() -> None:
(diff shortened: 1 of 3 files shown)


mrostan (approved):
nice, lgtm

PR #328, opened 2026-06-12
Validate self-upgrade image: registry allowlist, platform match, version floor

mrostan:
## What & why

A security review flagged the `/api/v1/upgrade` endpoint as a post-authentication RCE path: it accepted **any** image, so an authenticated caller could redeploy the agent with arbitrary code under the agent's cloud identity. The endpoint is reachable by the normal invoker and is enabled by default in the deployment terraform modules (`remote_upgradable = true`), so "disabled by default" did not hold.

This adds agent-side validation of the requested image. It does **not** rely on the backend — it independently constrains a direct, post-auth call to the endpoint (the attacker in the finding), and it complements what `monolith` already enforces when it builds the image.

## What it does

Validation runs at the single choke point in `Agent._perform_update`, so all four updaters (CloudRun, Azure, Lambda direct + CFN) are covered. Three checks:

1. **Registry/namespace allowlist** — the candidate must come from the same registry as the running image (and, on multi-tenant registries like docker.io, the same namespace); on a private registry such as ECR the account/host is the boundary. Removes the "arbitrary code" property. Widenable via `MCD_AGENT_UPGRADE_ALLOWED_REPOS`.
2. **Platform match** — rejects a recognized cross-platform swap (`cloudrun`↔`azure`) that would fail to boot. Availability/foot-gun guard.
3. **Version floor** — rejects downgrades (blocks downgrade-to-known-vulnerable); unversioned tags (`latest`, digest-only) are rejected by default. Pre-release ordering follows PEP 440 (`1.8.6rc1234 < 1.8.6`).

Parsing is registry-confusion resistant (anchored host parse), normalizes the ECR region wildcard (`*`), and strips the Azure `DOCKER|` linuxFxVersion prefix.

Verified against the authoritative backend path (`monolith/service/agents/agent.py::_get_image`): it never sends `latest`, forces the platform variant to match the agent, and sends AWS images as plain-semver ECR URIs with the `*` region wildcard — all of which the validator accepts.

## Testing

- New `tests/test_upgrade_validation.py` — parsing, allowlist bypass matrix (registry confusion, ECR account/wildcard, namespace lookalikes), platform match, version floor (rc/final ordering, downgrade, unversioned, dev `local`).
- Full suite: 1302 passed. `black --check` clean, `pyright` 0 errors.

## Notes for reviewers

- This does not change the deployment modules' `remote_upgradable` default or the Azure Contributor role grant — those live in the `terraform-*` repos and are tracked separately.
- Signature/attestation verification was discussed and intentionally **not** included.

### DIFF
--- a/apollo/agent/agent.py
+++ b/apollo/agent/agent.py
@@ -38,6 +38,10 @@ from apollo.common.agent.models import (
 from apollo.agent.proxy_client_factory import ProxyClientFactory
 from apollo.agent.settings import VERSION, BUILD_NUMBER
 from apollo.agent.updater import AgentUpdater
+from apollo.agent.upgrade_validation import (
+    get_configured_allowed_repos,
+    validate_upgrade_image,
+)
 from apollo.agent.utils import AgentUtils
 from apollo.credentials.factory import (
     CredentialsFactory,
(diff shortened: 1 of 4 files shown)


benitezfede (approved):
LGTM

PR #329, opened 2026-06-12
fix(azure): drop unused SignalR extension to clear MessagePack CVE-2026-48109

mrostan:
## What & why

Today's vuln scan flagged a new fixable **HIGH** on `latest-azure` only:
`CVE-2026-48109` — `MessagePack 2.5.192` (out-of-bounds read / DoS in the LZ4 path).

The vulnerable `MessagePack.dll` isn't one of our deps — it's baked into Microsoft's
Azure Functions extension bundle. Its **sole consumer is the SignalR Service extension**
(verified against the bundle's `function.deps.json` reverse-dependency graph). Our Azure
Function app only uses **HTTP + Durable Functions** triggers — never SignalR — so the whole
SignalR assembly cluster is dead weight and MessagePack is reachable code we don't need.

## Why delete, not patch the DLL

Docker Scout reports a .NET package only when **both** its `deps.json` entry **and** its DLL
exist on disk. I verified locally:
- **Overwriting** `MessagePack.dll` with a patched 2.5.301 build → CVE **still reported**
  (Scout reads the version from the manifest, not the binary).
- **Deleting** the DLL → CVE **cleared**.

So we delete the SignalR-cluster assemblies and de-register the extension. As a bonus this
removes ~6 unused libraries from the image's attack surface, retiring a recurring class of
bundle-CVE findings.

## Changes

- **`resources/azure/prune_signalr_extension.py`** — deletes the SignalR-cluster DLLs
  (`MessagePack(.Annotations)`, `Microsoft.Azure.SignalR.*`, `…SignalRService`) from every
  bundle bin dir and de-registers SignalR from each `extensions.json`. **Fails the build
  loudly** if its targets aren't found (a future bundle restructure surfaces as a build error,
  not a silent no-op that lets the CVE back in). Bundle version dir is globbed → survives the
  pending python3.13 base bump.
- **`Dockerfile.azure`** — the azure target extracted into its own file (it builds `FROM`
  Microsoft's Functions base and shares nothing with the `python:3.12-slim` `base` stage), with
  the prune wired in. `Dockerfile` keeps a pointer note; CI now uses `-f Dockerfile.azure`.
- **`apollo/interfaces/azure/README.md`** — fixed the stale `Dockerfile_az` reference; documented
  the smoke test.
- **`resources/azure/durable_smoke_test.sh`** — boots the built image against Azurite and runs a
  Durable Functions orchestration end-to-end.

## Validation (local)

- Scout on the built image: **0 MessagePack vulnerabilities**; 0 SignalR/MessagePack DLLs remain.
- Smoke test **passes**: host boots with no SignalR/MessagePack load errors, DurableTask loads on
  the AzureStorage backend, orchestration reaches `Completed`.

## Notes

- **Image size is unchanged** — deletion clears the *scan* (scanner reads the merged FS) but the
  base-layer bytes remain; reclaiming them would require flattening, which isn't worth the
  fragility here. The win is vuln/attack-surface reduction, not size.
- **Smoke test is intentionally not in CI.** The prune's build-time fail-loud runs in both the
  merge-queue and dev-merge builds, and dev deployment exercises Durable Functions for real — so
  both regression modes are already covered. Run the smoke test locally before merging to dev and
  after any base-image change.


mrostan:
Obsolete: MS shipped the fix upstream. Extension bundle is now 4.37.1 with MessagePack 2.5.301 (patched, CVE-2026-48109 fixed in 2.5.301) on both 4-python3.13 and 4-python3.14 base images. Base tags are unpinned, so the next azure build picks up the fixed bundle automatically — the SignalR prune is no longer needed.

PR #331, opened 2026-06-15
fix(deps): bump agent-base to v0.0.3 — redact credentials from error responses

mrostan:
## What & why

A security review found (**HIGH, CVSS 7.5** — unintentional inclusion of sensitive data in network messages):

> Any exceptions that occur in executing a command are sent to the MCD SaaS … If these exception tracebacks contain sensitive data (e.g. a database connect string), that entire string, including credentials, will be sent to the SaaS [and] logged in the customer account.

Error responses serialize the exception into `__mcd_error__` / `__mcd_exception__` / `__mcd_stack_trace__` with no redaction, so any credential that reaches an exception message or traceback is sent to the SaaS verbatim. This is most acute for the pyodbc clients (SQL Server, Azure SQL, MS Fabric), which assemble credentials into a single ODBC connection string (`UID=...;PWD=...`); kwargs-based clients like Postgres don't carry the secret as one string. The reviewer flagged this for connect-string-based databases. This change strips connect-string-shaped secrets from error responses regardless of the exact path.

**Caveat on reproducibility:** the leak is conditional on the driver (or a wrapper) actually surfacing the connection string in its exception. We could not reproduce it with our current drivers — psycopg2 is called with kwargs and never includes the password in its exception/traceback, and pyodbc did not echo the connection string on the failure paths we could test. So this lands as **defense-in-depth** against that class of leak rather than a fix for a confirmed live exposure.

The fix lives in agent-base ([PR #8](https://github.com/monte-carlo-data/agent-base/pull/8), released as **v0.0.3**): `_response_for_error()` now strips secret *values* inline (URL userinfo, `key=value`/JSON connect-string pairs, `Bearer`/`Basic` headers, bare high-entropy tokens) while leaving the surrounding message intact, so errors stay actionable.

## This PR

- Bumps the `agent-base` pin `v0.0.2` → `v0.0.3` (`requirements.in` + regenerated `requirements.txt`; only the pin changed, no transitive churn).
- Adds a regression test on the SQL Server (pyodbc) client — the structurally-exposed one — that fails the connection with an exception carrying the ODBC connection string (`PWD=...`) and asserts the password is stripped from the response while the server/error context survive. **Fails on v0.0.2, passes on v0.0.3.**

The existing `*CredentialSafetyTests` (actionable-and-safe contract) continue to pass — the inline redactor was chosen specifically so it doesn't clobber messages like `"authentication failed for user"`.

## Tests

Full suite green against the real v0.0.3 dependency. `black` + `pyright` clean.

## Checklist
- [x] Tests added/updated
- [x] `black` + `pyright` pass
- [na] Docs — dependency bump + internal redaction, no user-facing docs

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -1,4 +1,4 @@
-agent-base @ git+https://github.com/monte-carlo-data/agent-base.git@v0.0.2
+agent-base @ git+https://github.com/monte-carlo-data/agent-base.git@v0.0.3
 
 azure-identity==1.18.0
 azure-core==1.38.2  # YET-463
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@
 #
 #    pip-compile --output-file=requirements.txt --strip-extras requirements.in
 #
-agent-base @ git+https://github.com/monte-carlo-data/agent-base.git@v0.0.2
+agent-base @ git+https://github.com/monte-carlo-data/agent-base.git@v0.0.3
     # via -r requirements.in
 asn1crypto==1.5.1
     # via snowflake-connector-python
--- a/tests/test_sql_server_client.py
+++ b/tests/test_sql_server_client.py
@@ -1,4 +1,5 @@
 import datetime
+import json
 from typing import (
     Iterable,
     List,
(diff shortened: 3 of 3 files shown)


benitezfede (approved):
LGTM

PR #332, opened 2026-06-15
refactor(cloudrun): pin stdout StructuredLogHandler for logging

mrostan:
## What & why

A customer security review flagged (MEDIUM, CVSS 5.5) that Apollo's Cloud Run logs go through a buffered, API-based handler whose in-process gRPC buffer an attacker with RCE could drain before ingestion — deleting evidence of compromise.

The premise is **incorrect**. On Cloud Run, `Client.setup_logging()` auto-selects the handler from the detected resource type, and for `cloud_run_revision` the library already returns the **stdout-based `StructuredLogHandler`**, not the buffered `CloudLoggingHandler` (see `get_default_handler()` in `google/cloud/logging_v2/client.py:390-391`, `google-cloud-logging==3.12.1`). Logs are written to stdout and ingested **out-of-process** by Cloud Run — there is no in-process log buffer to tamper with.

This PR makes that explicit so the finding can't be re-raised by a future reviewer (or silently regress):

- Extracts logging setup into `setup_cloud_run_logging()` (own module, so it's testable without `main.py`'s import-time `Client()` + Flask construction).
- Pins `StructuredLogHandler` explicitly instead of relying on environment auto-detection. **Behavior-preserving** on Cloud Run; also removes the theoretical fallback to the buffered handler if resource detection ever failed.
- In-code docstring documents the stdout / out-of-process log path.

## Testing

- New `tests/test_cloud_run_logging_setup.py` (4 cases): asserts the installed handler **is** a stdout `StructuredLogHandler`/`StreamHandler`, **is not** a `CloudLoggingHandler`, has no `transport`, sources `project_id` from the client, and honors the debug log-level flag.
- Test-first proof: temporarily swapping the helper back to `CloudLoggingHandler` makes all 4 tests fail; reverting makes them pass.
- Full run green (new + `test_cloud_run_platform.py` + `test_cloud_run_updater.py` = 7 passed); `black` + `pyright` pre-commit hooks pass.

## Checklist

- [x] Tests added/updated
- [x] Lint/format (black) and type checks (pyright) pass
- [na] Docs — no user-facing/config changes; rationale captured in code + this PR
- [na] DB migrations
- [x] Behavior-preserving in production (same handler the library already selected on Cloud Run)

### DIFF
--- a/apollo/interfaces/cloudrun/logging_setup.py
+++ b/apollo/interfaces/cloudrun/logging_setup.py
@@ -1,0 +1,30 @@
+import logging
+
+from google.cloud.logging import Client
+from google.cloud.logging.handlers import StructuredLogHandler, setup_logging
+
+
+def setup_cloud_run_logging(client: Client, is_debug_log: bool) -> None:
+    """Install the stdout-based ``StructuredLogHandler`` for Cloud Run.
+
+    We pin ``StructuredLogHandler`` explicitly rather than relying on
+    ``Client.setup_logging()``'s environment auto-detection. For the
+    ``cloud_run_revision`` resource type the library already selects this same
+    handler (see ``get_default_handler()`` in
+    ``google/cloud/logging_v2/client.py``), so this is behavior-preserving.
+
+    Pinning it makes the log path unambiguous: records are written to stdout and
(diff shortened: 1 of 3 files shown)


benitezfede (approved):
LGTM

PR #333, opened 2026-06-15
fix(security): delete and lock down credential temp files (TLS keys/CA, SSH key)

mrostan:
## What

Closes a customer security-review finding: **TLS client keys / CA bundles downloaded from a secrets manager were never deleted and were written world-readable**, so they persisted for the lifetime of the container and were reachable by anyone with even temporary container access.

This PR makes credential temp files (a) owner-only, (b) uniquely-named per client, and (c) deleted when no longer needed — across the CTP pipeline, the proxy-client-internal writers, and the git integration.

## How

**Permissions + unique paths** — `SslOptions.write_ca_data_to_temp_file` now mints a fresh `mkstemp` path per call (always `0o600`, unguessable name) instead of a deterministic `/tmp/{hash}` path written with `open()` (`0o644`). Unique paths mean one client closing can never delete a CA file still in use by another client with the same CA data. Covers db2, teradata, aws, and the CTP `resolve_ssl_options` transform.

**Proxy-client cleanup lifecycle** — `BaseProxyClient.close()` is a template method: it runs the subclass teardown (`_close_client()`) and then deletes any temp files registered via `register_temp_files()`, in a `finally` so cleanup runs even if the connection teardown raises. Subclasses override `_close_client()` (not `close()`).

**CTP pipeline** — the file-writing transforms (`tmp_file_write`, `write_ini_file`, `resolve_ssl_options`) record their paths in `PipelineState.temp_files`; the pipeline surfaces them via an optional out-param and the proxy-client factory registers them on the constructed client. Resolves the long-standing TODO in `tmp_file_write`.

**Proxy-client-internal writers** — db2/teradata/aws register their legacy-path CA files for cleanup. `BaseDbProxyClient.get_cert_path` (used by **Presto** and **MySQL**) is now an instance method that registers the cert it downloads, so those connectors no longer leak it.

**Git** — `GitCloneClient` now uses **unique per-client paths** for both the repo dir and the SSH private key (`mkstemp`, `0o600`) instead of the fixed, shared, predictable `/tmp/repo` and `/tmp/mcd_rsa`. They are deleted in a `finally` (even on clone/read failure). This prevents leaks *and* the cross-request collision possible on long-lived concurrent runtimes (EKS/hermes, Azure Functions), where one request could otherwise overwrite or delete another's key mid-clone.

Oracle needed no change — it already uses `NamedTemporaryFile` (`0o600`) and deletes its cert/key in a `finally`.

## Review

Reviewed with the multi-agent `/code-review` (security, correctness, architecture, testing, devdocs). Findings F1–F7 fixed/resolved; two reviewer findings rejected as verified false positives (see the review summary comment). `db/CLAUDE.md` and `ctp/CLAUDE.md` updated to document the `_close_client` override contract and the temp-file registration obligation.

## Testing

- New: `tests/test_ssl_options.py` (owner-only + unique-path-per-call), `tests/test_proxy_client_temp_file_cleanup.py` (cleanup incl. cleanup-runs-when-`_close_client`-raises, and `get_cert_path` registration), `tests/ctp/test_temp_file_cleanup.py` (resolve surfaces paths).
- Updated: `tests/test_git_client.py` (unique paths; key deleted after clone and on error), `tests/test_db2_client.py` / `tests/test_teradata_client.py` / `tests/test_base_aws_proxy_client.py` (CA file registered + cleaned up), transform tests assert `state.temp_files`.
- A local docker-compose TLS-Postgres integration test was used to confirm a real `verify-ca` handshake uses the materialized CA file and that it's deleted on close (not committed — kept local).
- Full suite: **1350 passed** on Python 3.13; `black` + `pyright` clean.

## Checklist

- [x] Tests added/updated and passing locally
- [x] `black` + `pyright` pre-commit hooks pass
- [x] Ran `/code-review`
- [na] Schema/migration changes
- [x] Docs updated — `db/CLAUDE.md` and `ctp/CLAUDE.md` document the new cleanup contracts
- [x] Backwards compatible — CTP `resolve()`/`execute()` retain their existing return contract; the temp-files out-param is optional


mrostan:
## 🔍 Code review — credential temp-file cleanup

Independent multi-agent review (security, correctness, architecture, testing, devdocs — two passes each) against `90bb5c1`. Fixes pushed in `a8e4379` (F1–F5) and `69b472c` (F6).

| ID | Sev | Finding | Status |
|----|-----|---------|--------|
| F1 | ISSUE | Presto & MySQL leaked the CA cert from `get_cert_path` (never registered) | ✅ fixed — `get_cert_path` is now an instance method that self-registers |
| F2 | ISSUE | Deterministic/fixed CA temp paths → one client's close deletes a file still in use by another | ✅ fixed — unique `mkstemp` path per call (db2/teradata/aws/resolve_ssl_options) |
| F6 | ISSUE | Git repo dir & SSH key used fixed shared `/tmp` paths (concurrency collision on EKS/Azure; predictable key) | ✅ fixed — unique per-client paths; key via `mkstemp` |
| F4 | SUGGESTION | Missing registration/temp_files test assertions (teradata, aws, transforms) | ✅ fixed |
| F5 | SUGGESTION | `_close_client` contract + temp-file obligation undocumented | ✅ fixed — db/CLAUDE.md + ctp/CLAUDE.md |
| F7 | SUGGESTION | AWS fixed CA bundle path | ✅ resolved — subsumed by F2 |
| F8 | NIT | `_remove_temp_files` logs the path via `exc_info` | ⏸️ deferred (path not secret; pre-existing) |
| F9 | NIT | `_temp_files` not init in `__init__` | ⏸️ deferred (`getattr` guard makes current code correct) |

**Rejected (2 false positives, verified):**
- `tmp_file_write` "briefly world-readable" — `NamedTemporaryFile` creates `0o600` via `mkstemp`; no window.
- F3 git key-deletion tests "vacuous" — verified empirically that `write_key` creates the file and the assertion fails if `delete_key` is a no-op. Tests are load-bearing; the reviewers' premise was wrong.

Also: removed the now-redundant `write_owner_only` helper (unique `mkstemp` paths make its atomic-replace unnecessary) and cleaned up dead imports.

**Verification:** full suite **1350 passed** on Python 3.13; `black` + `pyright` clean.

apollo/agent/proxy_client_factory.py
```diff
@@ -530,17 +530,30 @@ def _create_proxy_client(
...
+                connection_type,
+                credentials,
+                context={"platform": platform},
+                temp_files=temp_files,
             )
         factory_method = _CLIENT_FACTORY_MAPPING.get(connection_type)
         if factory_method:
-            return factory_method(credentials, platform=platform)
+            client = factory_method(credentials, platform=platform)
+            client.register_temp_files(temp_files)
```
sassmith: **Cleanup may not run for non-cached `ctp_config` clients on the success path.**

When `ctp_config is not None`, `get_proxy_client` bypasses the cache and builds a fresh client every call. But `Agent._execute_with_client`'s `finally` only calls `client.close()` when `operation.skip_cache` is true, and on error it calls `dispose_proxy_client` — which is a cache-keyed no-op for a client that was never cached. `ctp_config` and `operation.skip_cache` are independent request inputs, so a custom-CTP call with `skip_cache=False` produces a client that registers temp files here but is never explicitly closed on the happy path. Cleanup then falls back to `__del__` (only `BaseDbProxyClient` defines one) / GC — which is exactly the GC-dependent behavior this PR is trying to move away from, and non-DB proxy clients have no `__del__` at all.

So for the custom-CTP path, the cert/key/ini files registered here can still linger. One option is to close any non-cached client in `_execute_with_client`'s `finally`, e.g. gate on `operation.skip_cache or ctp_config is not None` rather than `skip_cache` alone.

apollo/agent/proxy_client_factory.py
```diff
@@ -530,17 +530,30 @@ def _create_proxy_client(
...
-                connection_type, credentials, context={"platform": platform}
+                connection_type,
+                credentials,
+                context={"platform": platform},
+                temp_files=temp_files,
             )
         factory_method = _CLIENT_FACTORY_MAPPING.get(connection_type)
         if factory_method:
-            return factory_method(credentials, platform=platform)
+            client = factory_method(credentials, platform=platform)
```
sassmith: **Temp files leak if construction fails after the pipeline materializes them.**

By this point `resolve()/resolve_custom()` have already written the cert/key/ini files and appended their paths to `temp_files`. If `factory_method(...)` raises here (bad driver args, connection setup that happens in `__init__`, etc.), those files were never registered on any client, so nothing ever deletes them — the materialized TLS key/CA persists for the container lifetime, which is the leak this PR is closing.

Consider wrapping the construction so `temp_files` are `os.unlink`ed on failure before re-raising (applies to all three construction branches — the factory_method branch and both custom-connector branches).

sassmith (commented):
A couple of lifecycle observations on the temp-file cleanup — both about cases where `close()` (and therefore `_remove_temp_files`) may never run for files registered here in the factory. Left them inline. Not blocking; flagging because they're in the path this PR is specifically hardening.

sassmith (approved):

apollo/agent/proxy_client_factory.py
mrostan: Fixed in f844459. `_create_proxy_client` now wraps the resolve + all three construction branches in a `try/except BaseException` that `os.unlink`s the materialized `temp_files` before re-raising, so a constructor that fails after the pipeline wrote the cert/key/ini files no longer leaks them. Added `test_temp_files_cleaned_up_when_construction_fails` (postgres + `ca_data`, `psycopg2.connect` raises → asserts no `/tmp/*_ssl_ca.pem` left behind).

apollo/agent/proxy_client_factory.py
mrostan: Fixed in f844459. The `finally` now closes any non-cached client: `non_cached = operation.skip_cache or ctp_config is not None` → `client.close()`, on both the success and error paths. Went slightly beyond the suggestion since the error path leaked too (for an uncached `ctp_config` client, `dispose_proxy_client` is a cache-keyed no-op) — closing covers both. Cached-client behavior is unchanged. Added `test_ctp_config_client_closed_even_when_not_skip_cache`.

PR #336, opened 2026-06-18
fix(docker): Python 3.12 → 3.13 base image + numpy 2 lockfiles + retire telnet (VULN-1230)

mrostan:
## What & why

Closes **VULN-1230** — bumps `apollo-agent` Docker base images **Python 3.12 → 3.13.13** to clear base-image CVEs (expat CVE-2026-7210, tarfile CVE-2025-13462). **Supersedes #326** (Dockerfile-only, red CI). 3.13 is the newest **GA** Python across all platforms we ship — Azure Functions is the binding constraint (3.14 is preview-only there).

## Changes

- **Base image + toolchain → 3.13**: Dockerfile (all stages), CircleCI lint/docs builders, `setup.py`, docs.
- **Retire telnet validator** (`telnetlib` removed in 3.13, PEP 594): the `/test/network/telnet` endpoint + `validate_telnet_connection` are kept for FE compatibility but now delegate to the TCP-open probe (through the SSRF guard). FE will drop the call separately.
- **Lockfiles regenerated under 3.13 — numpy 1→2 migration**: `numpy>=2.1.0` (first cp313 wheels) / pandas 2.2.3; `databricks-sql-connector 4.0.0→4.2.5` (highest 4.x still compatible with impyla's `thrift==0.16.0` pin); `psycopg2-binary→2.9.12`, `pyodbc→5.3.0`; transitive cp313 floors (`google-crc32c`, `lz4`, `markupsafe`). Platform lockfiles stay consistent via their `-c requirements.txt` constraint.
- **Azure Functions worker (py3.13 startup) — 3 fixes**: `azure-monitor-opentelemetry 1.3.0→1.8.8` (the old one imports `pkg_resources` and crashes on 3.13); `google-cloud-storage 2.10.0→2.19.0` (its legacy setuptools `google` namespace corrupts under 3.13 → `KeyError: 'google'`); install app deps into `.python_packages` so the host's dependency isolation keeps our protobuf off the worker's `sys.path` (was a SIGSEGV).
- **Logging redaction fix** (`agent-base @v0.0.4`, [agent-base#9](https://github.com/monte-carlo-data/agent-base/pull/9)): `RedactFormatterWrapper` now redacts the *formatted* message, not the format string — the old order raised `TypeError` mid-log and broke Azure credential/storage ops.
- **cryptography 48.0.0→48.0.1** (GHSA-537c-gmf6-5ccf, bundled-OpenSSL OOB read; folds in #337).
[...]

PR #337, opened 2026-06-22
fix(deps): bump cryptography 48.0.0 -> 48.0.1 (GHSA-537c-gmf6-5ccf)

mrostan:
## What

Bumps `cryptography` from `48.0.0` to `48.0.1` to clear **GHSA-537c-gmf6-5ccf** (HIGH, CVSS 7.5), surfaced in the daily `#collection-agent-vuln-scans` report (2026-06-21) as a *shared* finding across all agent image tags.

The 48.0.0 PyPI wheels bundle a statically-linked OpenSSL with an out-of-bounds read (per OpenSSL's 2026-06-09 advisory). pyca/cryptography rebuilt its wheels against the patched OpenSSL in 48.0.1.

## Changes

- `requirements.in`: raise the floor `cryptography>=46.0.5` → `>=48.0.1`, documenting the advisory inline (prevents regression to a vulnerable wheel).
- `requirements.txt`: re-pin `cryptography==48.0.1`.

No other lockfile changes: on `main`, cryptography is only a direct entry in the base `requirements.txt` — the azure/cloudrun/lambda/dev lockfiles install it transitively via `-c requirements.txt`, so this single bump covers every tag.

## Verification

Regenerated `requirements.txt` with `pip-compile --strip-extras` in a clean **Python 3.12** venv (pip-tools 7.5.3, matching the lockfile header). Result was a **single-line diff** (`48.0.0` → `48.0.1`) with no transitive churn and no jump to the newly-released `49.0.0` — the bump is surgical and the lockfile is reproducible. CI image build + scan is the final gate.

## Key Decisions

- **Patch bump, not latest.** `49.0.0` is now out, but a CVE fix shouldn't drag in a major version. Re-pinning the lockfile to `48.0.1` before recompiling (no `--upgrade`) keeps pip-compile pinned to the patched version only.
- **Based on `main` (Python 3.12), not the in-flight 3.13 migration branch.** The 3.13 work (VULN-1230) also pins `cryptography==48.0.0` in its azure lockfile and will need the same 48.0.1 bump independently — this PR does not reach it.
- Other open scan findings (MessagePack `CVE-2026-48109` azure-only; Debian perl/xvfb/libnss3/libhttp-daemon CVEs) are either already in flight or have no upstream fix, and are out of scope here.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -9,7 +9,7 @@ brotli==1.2.0  # Upgrade transient dependency of flask-compress to address SEC-1
 cachetools==5.3.3  # Runtime dep for apollo/credentials/external_credentials_cache.py; pinned here so it shows up in install_requires for downstream consumers
 cerberus>=1.3.5
 clickhouse-connect==0.8.18
-cryptography>=46.0.5  # CVE-2026-26007
+cryptography>=48.0.1  # CVE-2026-26007; >=48.0.1 for GHSA-537c-gmf6-5ccf (bundled OpenSSL OOB read, fixed in 48.0.1)
 databricks-sql-connector==4.0.0
 databricks-sdk==0.39.0
 duckdb==1.1.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -74,7 +74,7 @@ click==8.1.7
     #   presto-python-client
 clickhouse-connect==0.8.18
     # via -r requirements.in
-cryptography==48.0.0
+cryptography==48.0.1
     # via
     #   -r requirements.in
     #   azure-identity


mrostan:
Superseded by #336 (VULN-1230 py3.13 migration), which folds the cryptography 48.0.1 bump (GHSA-537c-gmf6-5ccf) in directly as commit 8474776 — same `requirements.in` floor (`>=48.0.1`) and re-pinned base + azure lockfiles. Closing to avoid a redundant/competing change; the fix ships with #336.

PR #338, opened 2026-06-23
fix(db2): accept ODBC-spelled credential keys in CTP mapper

mrostan:
## What
Adding Db2 connections to agents was failing with this error:
image

The Db2 CTP mapper only read the **flat** credential spellings (`raw.host`, `raw.user`, `raw.password`), but the data-collector agent path (`clients/plugins/plugin_db2.py`) pre-shapes credentials into `connect_args` using IBM **ODBC** key names (`hostname`/`uid`/`pwd`/`database`). The registry unwraps `connect_args` before running the pipeline, so `raw.host` was undefined and integration validation failed with **`'host' is undefined`**.

This fix adds the ODBC spellings as fallbacks in the mapper, mirroring how `DATABASE` already handled `db_name` → `database`. The misleading schema comment (which had the DC path's spellings backwards) is also corrected.

## Why it shipped broken

The existing `tests/ctp/test_db2_ctp.py` tests only exercised flat creds via `CtpPipeline().execute()` **directly** — they never went through the registry's `connect_args` unwrap path that the real agent uses. So the mismatch was never caught.

## Repro

Add a Db2 integration via an agent → "Test Db2 integration" → "Connect to Db2" fails with Cause `'host' is undefined`. Confirmed not a monolith-django or frontend issue (both consistently send `host`); the rename to ODBC keys happens in data-collector's pre-shape, and the agent's CTP mapper couldn't find them.

## Tests

- 3 ODBC-alias tests (`hostname`/`uid`/`pwd`) matching the existing alias-test style.
- 1 end-to-end `test_dc_preshaped_connect_args_path` resolving through `CtpRegistry.resolve` — replicates `plugin_db2.py` exactly (the true regression test).
- **Test-first verified:** all 4 new tests failed with the undefined error before the fix.
- After fix: 17/17 in the db2 file; **522/522** across `tests/ctp` + `tests/test_db2_client.py`. `black --check` clean, `pyright` 0 errors.

## Notes

- Fixed at the CTP mapper layer (accept both spellings) — the robust layer. Alternative would be to stop renaming keys in `plugin_db2.py` (data-collector), but that's riskier.

### DIFF
--- a/apollo/integrations/ctp/defaults/db2.py
+++ b/apollo/integrations/ctp/defaults/db2.py
@@ -21,9 +21,10 @@ class Db2ClientArgs(TypedDict):
     SSLServerCertificate: NotRequired[str]  # path to CA PEM file
 
 
-# DB2 self-hosted credentials schema mirrors the docs accordion. The CTP also
-# accepts host/user/password/db_name as alternate spellings on the DC pre-shape
-# path, but customer self-hosted JSON is expected to follow the docs.
+# DB2 self-hosted credentials schema mirrors the docs accordion: the DC pre-shape
+# path (clients/plugins/plugin_db2.py) sends connect_args with IBM ODBC key names
+# (hostname/uid/pwd/database). The mapper below also accepts the flat spellings
+# (host/user/password/db_name) so direct flat credentials resolve too.
 DB2_CREDENTIALS_SCHEMA = {
     "connect_args": {
         "type": "dict",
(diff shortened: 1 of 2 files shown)


mrostan:
Verified on a live Azure agent running this build (via `dev`): adding a Db2 connection now succeeds — the "'host' is undefined" validation error is resolved.

sassmith (approved):

benitezfede (approved):
LGTM

PR #345, opened 2026-07-03
YET-1640: log hard agent command failures at ERROR

squarejay:
## Changes
Elevate hard, data-dropping agent command failures from INFO to ERROR so they are alertable — without leaking sensitive data.

- `evaluation_utils.py`: method-invocation failures now log via `logger.exception` (captures the traceback) with only the exception *type* name. The raw exception string — which can carry presigned artifact URLs, tokens, or HTTP response bodies — is no longer interpolated into the message, because the agent's log redactor is key-based and the agent runs in customer infra.
- Method-resolution failures (`Failed to resolve method ...` — the Nutrien incident path) now log at ERROR before raising.

## Testing
- `AgentCommandFailureLoggingTests`: asserts ERROR level + stable `"Failed to resolve method"` substring on resolution failure; ERROR + `exc_info` + method name present + secret NOT leaked + re-raise on invocation failure.
- `black --check` clean, `pyright` 0 errors, `pytest tests/test_agent_commands.py` 7 passed.

## Context
Prevention/observability follow-up to YET-1618 (dbt artifact-fetch broke silently ~3 weeks across Salesforce/Nutrien). Part 1 of 4 for YET-1640. Defense-in-depth: for agent-proxy customers the agent runs in customer infra, so these ERROR logs mainly aid investigation; the load-bearing detection is a monolith-side per-account metric + monitor (later YET-1640 phases).

### DIFF
--- a/apollo/agent/evaluation_utils.py
+++ b/apollo/agent/evaluation_utils.py
@@ -159,10 +159,12 @@ class AgentEvaluationUtils:
                 )
             except Exception as ex:
                 duration_s = time.monotonic() - t0
-                logger.info(
-                    f"Command failed, method={command.method}, "
-                    f"target={type(target).__name__}, duration_s={duration_s:.3f}, "
-                    f"error={ex}"
+                logger.exception(
+                    "Command failed, method=%s, target=%s, duration_s=%.3f, error_type=%s",
+                    command.method,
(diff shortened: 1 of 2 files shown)


mrostan (approved):
lgtm

PR #347, opened 2026-07-06
fix(redshift): accept IAM-federated secrets in self-hosted credential validation [SUP-532] (dev)

gfitzpatrick-mcd:
## What

Redshift self-hosted credential validation rejected **IAM-federated** (Connection Auth Rules) secrets: `REDSHIFT_CREDENTIALS_SCHEMA` required `connect_args.{host, dbname, password}`. Federated secrets have **no static password** (temporary creds are minted at connect time via `GetClusterCredentials`) and use `db_name`, not `dbname` — so customers had to add a dummy password and duplicate `db_name`/`dbname` keys just to pass validation.

This converts the Redshift raw-credentials schema to a cerberus `oneof_schema` with a **password variant** and a **federated variant** (`cluster_identifier`/`db_user`/`aws_region`, no password) over a shared base — mirroring the existing Snowflake multi-auth-mode pattern. `db_name`/`dbname`/`database` are accepted as aliases so any documented spelling validates.

## Key decisions

- **Fixed in the static schema, not the validator.** `validate_self_hosted_credentials` only receives `connection_type` (no `ctp_config`/auth rule), and neither transport passes it — so a schema-level fix (like Snowflake) is correct and avoids cross-transport plumbing.
- **db-name aliases kept optional** (`db_name`/`dbname`/`database`) rather than requiring one, to avoid rejecting existing `dbname`-keyed password secrets while accepting the documented `db_name`; a genuinely missing db name surfaces at connect time.
- **`host` kept required in both variants** — federated still connects to the cluster endpoint.

## Testing

- New endpoint tests: federated (no password) passes, password passes, neither-auth fails with a `connect_args` error.
- Existing redshift CTP + schema-registry tests green (45), endpoint suite 16 pass, ruff clean.

Linear: SUP-532 · Customer: ZD #28116 (BBC)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,14 +2,14 @@
 # Published as `<version>-system-base` so downstream consumers (e.g. hermes-agent)
 # can build their own venv against the same native libs without inheriting
 # apollo's pip-installed dependencies.
-FROM python:3.13.13-slim AS system-base
+FROM python:3.13.14-slim AS system-base
 
 ENV APP_HOME=/app
 WORKDIR $APP_HOME
 
 # Refresh apt index and upgrade base-image packages so OS-level security fixes
 # (glibc, openssh, nghttp2, etc.) land on every rebuild rather than waiting for
-# the upstream python:3.13.13-slim tag to be republished.
+# the upstream python:3.13.14-slim tag to be republished.
 RUN apt-get update && apt-get upgrade -y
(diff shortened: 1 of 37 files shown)


gfitzpatrick-mcd:
Superseded by a clean dev PR cherry-picked onto origin/dev (this one carried unrelated main→dev commits).

PR #350, opened 2026-07-13
Add Oracle thick mode support to the agent

mrostan:
## What

Adds Oracle **thick mode** (Oracle Instant Client) support to the agent. Some Oracle configurations require thick mode — e.g. accounts that only expose the legacy password verifier, which the pure-Python thin driver can't authenticate with.

## Design: agent-level env var (not per-connection)

python-oracledb thick mode is **process-global and one-way** (`init_oracle_client()`) — thin and thick connections can't coexist in one process (a thin→thick switch raises `DPY-2019`). So it's an **agent-level env var, `MCD_ORACLE_THICK_MODE`**, initialized once per process.

> Enable it only on an agent **dedicated** to thick-mode Oracle. A shared agent that also serves thin Oracle will hit `DPY-2019`.

## Changes

- **Dockerfile** — Instant Client Basic 23.8 in all three variants (`system-base`, `lambda`, `azure`); handles Debian trixie's `libaio1t64` (compat symlink) and the Lambda base's `ldconfig` path.
- **`oracle_proxy_client.py`** — read `MCD_ORACLE_THICK_MODE`; `init_oracle_client()` once (guarded by `is_thin_mode()`); reject `ssl_options` in thick mode (thin-only — fail rather than connect without the requested SSL).
- **`MCD_ADDITIONAL_ENV_VARS`** — parse a JSON-object env var at startup and inject its entries into the environment, so IaC can set agent env vars (incl. `MCD_ORACLE_THICK_MODE`) without adding a dedicated parameter each time. CloudFormation has no map parameter type; Terraform `jsonencode`s a map into the same var — one carrier, one code path.
- **AWS Lambda logging** — Lambda pre-installs a root log handler, making `basicConfig()` a no-op → INFO logs were dropped; set the root level explicitly.
- **Health `env`** — surface any `MCD_`-prefixed env var (not just the allowlist), skipping sensitive-looking names (`secret`/`pass`/`token`/`key`/`credential`) so e.g. `MCD_STORAGE_ACCESS_KEY` can't leak.

## No data-collector change

The agent self-configures via the env var; the DC keeps sending plain `connect_args`.

## Enabling from IaC (companion PRs)

These set the env var for each deploy path (via `MCD_ADDITIONAL_ENV_VARS`, or a native mechanism where one already exists):

- Terraform AWS — monte-carlo-data/terraform-aws-mcd-agent#25
- Terraform Azure — monte-carlo-data/terraform-azurerm-mcd-agent#17
- Terraform Google — monte-carlo-data/terraform-google-mcd-agent#13
- CloudFormation (Lambda + ECS) — monte-carlo-data/mcd-public-resources#71
- Helm / generic (uses existing `container.extraEnv`) — monte-carlo-data/hermes-agent#59

## Testing

Unit tests (env-var enable / skip-when-thick / thin default / ssl_context ignored / additional-env-vars parsing / health filtering). All three image variants built locally with `init_oracle_client()` verified; real Oracle connection verified via a throwaway test image.

## Notes

x86_64 only (single-arch CI builds); Instant Client Basic adds ~200MB per variant.

### DIFF
--- a/apollo/integrations/db/oracle_proxy_client.py
+++ b/apollo/integrations/db/oracle_proxy_client.py
@@ -134,21 +134,15 @@ class OracleProxyClient(BaseDbProxyClient):
                 1  # enable keep-alive and send packets every minute
             )
 
-        # Thick mode (Oracle Instant Client) is a process-global, one-way switch
-        # enabled via init_oracle_client(). It cannot coexist with thin
-        # connections in the same process — once a thin connection exists,
-        # enabling thick raises DPY-2019 — so it is an AGENT-LEVEL setting
-        # (MCD_ORACLE_THICK_MODE), not per-connection. Enable it only on an agent
-        # dedicated to thick-mode Oracle (e.g. an Oracle EBS agent). The Instant
-        # Client native libraries must be present on the image (see Dockerfile);
-        # the OS loader finds them via ldconfig.
+        # Thick mode (Oracle Instant Client) is process-global and one-way: once a
+        # thin connection exists it can't be enabled (DPY-2019). So it's an
+        # agent-level env var, not per-connection — enable it only on a dedicated
+        # thick-mode Oracle agent.
         thick_mode = _thick_mode_enabled()
         if thick_mode and oracledb.is_thin_mode():
             # Runs once per process, on the first Oracle connection. is_thin_mode()
             # flips to False after a successful init, so later connections skip it.
             oracledb.init_oracle_client()
-            # Prefix with the "oracle" connection type so log searches by
-            # connection type match (case-sensitive platforms miss "OracleDB").
             logger.info("oracle: thick mode initialized")
 
         # Handle SSL options for Oracle connections. Thick mode does not support
--- a/apollo/integrations/db/oracle_proxy_client.py
+++ b/apollo/integrations/db/oracle_proxy_client.py
@@ -125,9 +125,28 @@ class OracleProxyClient(BaseDbProxyClient):
                 1  # enable keep-alive and send packets every minute
             )
 
-        # Handle SSL options for Oracle connections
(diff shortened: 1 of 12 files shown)


apollo/integrations/db/oracle_proxy_client.py
```diff
@@ -125,9 +134,27 @@ def __init__(self, credentials: Optional[Dict], **kwargs: Any):
...
+        # Handle SSL options for Oracle connections. Thick mode does not support
+        # ssl_context (thin mode only), so they are mutually exclusive.
         ssl_options = SslOptions(**(credentials.get("ssl_options") or {}))
-        if ssl_context := create_oracle_ssl_context(ssl_options):
+        if thick_mode:
+            if not ssl_options.disabled and ssl_options.ca_data:
+                logger.warning(
+                    "Oracle SSL via ssl_context is only supported in thin mode; "
+                    "ignoring ssl_options because thick mode is enabled"
+                )
```
sassmith, line 156: Is it better to fail if someone tries thick mode with ssl_context than ignore? I guess it puts the burden on the db side to reject the request if it doesnt contain the cert that is expected. I am really just thinking from a troubleshooting perspective, I dont think this warning log will surface in connection validators but maybe the connection attempt failure message would be clear enough.

sassmith (approved):

apollo/integrations/db/oracle_proxy_client.py
mrostan: as discussed in Slack: https://montecarloai.slack.com/archives/C09PBS17QMR/p1784209666819549
I have changed this code to raise an error and will update the equivalent code in data-collector to do the same, follow up PR: https://github.com/monte-carlo-data/data-collector/pull/2510

PR #352, opened 2026-07-21
Support SSL (TLS + mTLS) for Oracle thick mode

mrostan:
## What

Adds **SSL (TLS + mTLS) support for Oracle thick mode**, replacing the previous behavior where thick + `ssl_options` raised. Also fixes a related pre-existing gap where the Oracle CTP dropped `ssl_options` for agent connections, so **no** SSL (thin or thick) actually worked over the agent.

> Note: an earlier revision of this PR tried to build the wallet in pure Python (`ewallet.p12` via the `connect()` `wallet_location` param). That approach does **not** establish trust — see below — and was reworked.

## Why it's non-trivial

Thick mode doesn't accept a Python `ssl.SSLContext` (thin-only); it validates the server certificate against an Oracle **wallet**. Verified the hard way against a real RDS Oracle TLS endpoint:

- Thick reads the server trust store **only from `WALLET_LOCATION` in `sqlnet.ora`** — the `connect()` `wallet_location` param does **not** drive server-cert trust.
- The wallet must be a **`cwallet.sso`** produced by Oracle's **`orapki`** tool. A `cryptography`-built PKCS#12 is opened but never honored as a trust anchor (ORA-29024).
- The wallet must exist at `WALLET_LOCATION` **when `init_oracle_client` runs**, or the wallet subsystem starts with no trust store and every TLS connect fails with ORA-29024.
- Oracle **caches the trust store after the first successful connection** for the process lifetime — it does not re-read the wallet per connect.
- Thick has no `@SECLEVEL` knob, so older RSA-kx ciphers (e.g. RDS Oracle's `AES256-GCM-SHA384`) must be listed explicitly in `SSL_CIPHER_SUITES`.
- **The Oracle CTP was dropping `ssl_options` over the agent.** The mapper only emitted `dsn/user/password/expire_time`, so the `ssl_options` block (with `ca_data`) never reached `OracleProxyClient`. Every SSL connection through the agent failed: thin connected to `tcps` with no `ssl_context` → `SSLV3_ALERT_HANDSHAKE_FAILURE` (RDS's RSA cipher rejected at the default security level); thick connected with no wallet → ORA-29024. Both reproduced live against RDS.

## How

- **New `apollo/integrations/db/oracle_client_config.py`** owns all Oracle driver-mode + TLS logic (thin `ssl.SSLContext` and thick wallet), keeping `oracle_proxy_client.py` to orchestration.
- **Wallet via orapki:** `create_oracle_thick_wallet(ssl_options)` runs `orapki wallet create -auto_login`, adds each CA from `ca_data` as a `-trusted_cert`, and (for **mTLS**) imports the client key+cert from `cert_data`/`key_data` via `import_pkcs12`.
- **Correct ordering:** `configure_thick_connection(ssl_options)` builds the wallet and writes `sqlnet.ora` (`SSL_CIPHER_SUITES` + `WALLET_LOCATION` + `SSL_SERVER_DN_MATCH`) **before** `init_oracle_client`, once per process, under a lock. Because trust is cached after the first connection, one process-wide wallet is built from the first connection's `ssl_options`; a thick agent supports a single Oracle TLS trust config and **fails loudly** (rather than a confusing ORA-29024) if a later connection needs a CA the process can't apply.
- **CTP `ssl_options` passthrough (the agent fix):** the Oracle CTP now carries `ssl_options` through into `connect_args` (added to the `OracleClientArgs` TypedDict + mapper; the sandboxed NativeEnvironment preserves the dict instead of stringifying it). `OracleProxyClient` **pops** `ssl_options` from `connect_args` (it is not an `oracledb.connect` arg) and falls back to the top-level sibling for direct/legacy callers. This makes both thin and thick SSL work over the agent.
- **Validation by default:** `SSL_SERVER_DN_MATCH` follows `ssl_options.verify_identity` (default on) — nothing hardcoded off; the CA chain is always validated in thick mode.
- **Ciphers:** default list = RSA-kx (RDS/EBS) + modern ECDHE; override via `MCD_ORACLE_SSL_CIPHER_SUITES`.
- **Dockerfile:** new `oracle-pki-builder` stage builds a stripped ~54 MB JRE (jlink: `java.base`, `java.naming`, `jdk.crypto.ec`) plus the `oraclepki`/`osdt_core`/`osdt_cert` jars, copied into the `system-base`, `lambda`, and `azure` images so orapki can build wallets at runtime. Built on Amazon Linux 2023 (oldest glibc) and verified to run on both the Debian and Lambda AL2023 bases.
- Thin path unchanged.

## Testing

- Unit tests: thick+TLS wires trust via `sqlnet.ora` (not the `connect()` param, not `ssl_context`); wallet written **before** `init_oracle_client`; orapki commands issued for one-way and mTLS; fail-loud on unappliable/differing CA; CTP carries `ssl_options` through as a native dict and `OracleProxyClient` pops it. Oracle + full CTP suites pass; `black` + `pyright` clean.
- **One-way TLS verified end-to-end against a real RDS Oracle** — both directly through `OracleProxyClient` and **through the AWS agent** (thick, `MCD_ORACLE_THICK_MODE=true`, `protocol=tcps`) after the CTP fix.
- Docker `system-base` image builds; orapki produces a `cwallet.sso` inside it.
- **mTLS** is built + unit-tested but **not** validated against a client-auth server (RDS can't require client certs; would need a local Oracle-mTLS instance).

## Follow-ups (separate PRs)

- data-collector (#2517): revert the thick+SSL raise and reuse `oracle_client_config` for its local-thick flow.
- Revert the "thick can't do SSL" wording in the docs (done on the docs branch).

Draft until reviewed.

### DIFF
--- a/apollo/integrations/db/oracle_client_config.py
+++ b/apollo/integrations/db/oracle_client_config.py
@@ -292,9 +292,10 @@ def configure_thick_connection(ssl_options: SslOptions) -> None:
     On the FIRST Oracle connection this builds the wallet from ``ssl_options`` and
     writes ``sqlnet.ora`` BEFORE ``init_oracle_client`` — the ordering Oracle
     requires for the wallet subsystem to start with the trust store present.
-    Because Oracle caches the trust store after the first connection, later calls
-    do nothing but warn if a connection asks for a TLS config that differs from
-    the one already established (a thick agent supports a single TLS trust config).
+    Because Oracle's trust store is process-global and frozen at the first
+    ``init_oracle_client``, a later connection that needs TLS trust the process
+    can't apply (initialized without a wallet, or with a different CA) is rejected
+    with a clear error rather than failing later with a confusing ORA-29024.
 
     The caller runs ``oracledb.connect`` after this returns.
     """
@@ -316,28 +317,37 @@ def configure_thick_connection(ssl_options: SslOptions) -> None:
                 + ("; TLS wallet configured" if _thick_wallet_dir else "")
             )
         else:
-            _warn_on_tls_config_change(ssl_options)
+            _require_compatible_established_tls(ssl_options)
 
 
-def _warn_on_tls_config_change(ssl_options: SslOptions) -> None:
-    """Warn when a later connection wants a TLS config the process can't apply.
+def _require_compatible_established_tls(ssl_options: SslOptions) -> None:
+    """Reject a connection whose TLS trust the process can't apply.
 
-    Thick trust is frozen after the first connection, so a differing CA (or TLS
-    requested when the agent was initialized without it) is silently ineffective;
-    surface it rather than let it fail confusingly.
+    Thick-mode trust is process-global and fixed at the first
+    ``init_oracle_client``. If a TLS connection arrives after the process was
+    initialized without a wallet, or with a different CA, its trust cannot be
+    applied — fail loudly with actionable guidance instead of letting Oracle
+    fail later with ORA-29024 (certificate validation failure).
     """
     wants_tls = bool(not ssl_options.disabled and ssl_options.ca_data)
     if not wants_tls:
         return
     if _thick_wallet_dir is None:
-        logger.warning(
-            "oracle: thick mode was initialized without TLS by the first connection; "
-            "TLS trust cannot be added later in the same process"
+        raise RuntimeError(
+            "Oracle thick mode was already initialized in this agent process "
+            "without an SSL trust store, so this SSL connection cannot be "
+            "established — thick-mode TLS trust is process-global and fixed at "
+            "first use. Restart the agent so its first Oracle connection uses "
+            "this SSL configuration, or use a dedicated agent for this SSL "
+            "integration."
         )
-    elif _ca_fingerprint(ssl_options) != _thick_ca_fingerprint:
-        logger.warning(
-            "oracle: thick mode uses the TLS trust established by the first connection; "
-            "this connection's differing CA data is ignored"
+    if _ca_fingerprint(ssl_options) != _thick_ca_fingerprint:
+        raise RuntimeError(
+            "Oracle thick mode was already initialized in this agent process "
+            "with a different SSL CA. A thick-mode agent supports a single "
+            "Oracle TLS trust configuration (process-global). Restart the agent "
+            "so it initializes with this CA, or use a dedicated agent for this "
+            "integration."
         )
 
 
--- a/apollo/integrations/db/oracle_client_config.py
+++ b/apollo/integrations/db/oracle_client_config.py
@@ -97,7 +97,9 @@ logger = logging.getLogger(__name__)
 _thick_lock = threading.Lock()
 _thick_config_dir: Optional[str] = None  # holds sqlnet.ora (passed to init)
 _thick_wallet_dir: Optional[str] = None  # the single TLS wallet, or None if no TLS
-_thick_ca_fingerprint: Optional[str] = None  # sha256 of the established ca_data
+# sha256 of the established TLS material (CA + client identity), so a later
+# connection needing a different CA *or* client certificate is detected.
+_thick_tls_fingerprint: Optional[str] = None
 
 
 def thick_mode_enabled() -> bool:
@@ -112,9 +114,11 @@ def _run_orapki(*args: str) -> None:
     """Run an ``orapki`` wallet command via the bundled minimal JRE.
 
     ``args`` are the orapki arguments (e.g. ``"wallet", "create", ...``). Raises
-    ``RuntimeError`` with orapki's output on failure. The command line (which
-    includes ``-pwd``) is deliberately kept out of the exception message so the
-    wallet password never lands in logs.
+    ``RuntimeError`` with orapki's output on failure. The command line itself is
+    kept out of the message, and any password values we passed (``-pwd`` /
+    ``-pkcs12pwd``) are scrubbed from orapki's stdout/stderr before they reach the
+    exception — which propagates to logs AND the agent's API error response, where
+    no keyword/entropy redaction would catch a bare CLI-flag value.
     """
     home = _oracle_pki_home()
     java_bin = os.path.join(home, "jre", "bin", "java")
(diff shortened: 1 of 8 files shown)


apollo/integrations/db/oracle_proxy_client.py
```diff
@@ -13,107 +10,18 @@
...
-# Agent-level thick-mode switch. Thick mode (Oracle Instant Client) is a
-# process-global, one-way setting that cannot coexist with thin connections in
-# the same process, so it is configured per-agent via this env var rather than
-# per-connection. Enable it only on an agent dedicated to thick-mode Oracle.
-_ENV_VAR_THICK_MODE = "MCD_ORACLE_THICK_MODE"
 
 logger = logging.getLogger(__name__)
 
 
-def _thick_mode_enabled() -> bool:
```
mrostan, line 29: moved to oracle_client_config dedicated file

mrostan:
<details>
<summary>Code Review — 14 findings (12 fixed, 1 skipped, 1 deferred)</summary>

**Reviewed:** 2026-07-21 UTC | **Reviewers:** security, correctness, testing, architecture, devdocs, coordinator | **Scope:** full PR

| F1 | ISSUE | orapki failure output could leak the wallet/PKCS12 password to logs + API response | ✅ Fixed — scrub `-pwd`/`-pkcs12pwd` values before raising |
| F2 | ISSUE | Wallet dir (key material) orphaned if sqlnet write/init fails after wallet build | ✅ Fixed — cleanup + reset state on failure |
| F3 | ISSUE | mTLS compat check hashed CA only, ignoring client identity | ✅ Fixed — fingerprint covers CA + cert/key |
| F4 | SUGGESTION | `_run_orapki` had zero real test coverage | ✅ Fixed — argv + redaction tests |
| F5 | SUGGESTION | `_write_thick_sqlnet` file content never verified | ✅ Fixed — real-content tests |
| F6 | SUGGESTION | Multi-CA bundle loop only tested with a single cert | ✅ Fixed — 2-cert bundle test |
| F7 | SUGGESTION | Same-CA reconnect success path untested | ✅ Fixed — reconnect-success test |
| F8 | SUGGESTION | `db/CLAUDE.md` didn't note the process-global wallet exception | ✅ Fixed |
| F9 | SUGGESTION | `ctp/CLAUDE.md` omitted Oracle + the nested-dict passthrough pattern | ✅ Fixed |
| F10 | SUGGESTION | Thick globals poked directly by tests | ✅ Fixed — added `_reset_for_testing()` (dataclass consolidation deferred) |
| F11 | SUGGESTION | `pop(...) or ...` conflated empty-dict with absent ssl_options | ✅ Fixed — explicit `is not None` |
| F12 | SUGGESTION | Wallet password visible via orapki argv (`/proc`) | ✅ Fixed — documented accepted risk (orapki has no stdin/pwfile) |
| F13 | SUGGESTION | jars/IC fetched without checksum verification | 🔜 Deferred — supply-chain hardening pass (matches existing IC pattern) |
| F14 | NIT | `ssl_options: NotRequired[Any]` typing | ⏭️ Skipped — `Any` is fine for an opaque passthrough |

👉 Fix commits: [48e2714](https://github.com/monte-carlo-data/apollo-agent/commit/48e2714) (F1–F5), [bf42636](https://github.com/monte-carlo-data/apollo-agent/commit/bf42636) (F6–F12)

</details>

apollo/integrations/db/oracle_client_config.py
```diff
@@ -0,0 +1,498 @@
...
+    connection's wallet identity instead of being rejected. The key passphrase is
+    intentionally excluded — the cert/key already identify the client, and hashing
+    a password here would be misread as (weak) password storage.
+
+    This is a change-detection fingerprint, not password storage, so SHA-256 is
+    the appropriate choice.
+    """
+    if ssl_options.disabled or not ssl_options.ca_data:
+        return None
+    material = "\0".join(
```
sassmith, line 341: Question on the fingerprint scope: `SSL_SERVER_DN_MATCH` is written once from the first connection's `verify_identity` and then frozen process-wide, but this fingerprint only covers `ca_data`/`cert_data`/`key_data` — not `verify_identity`. So two connections with the same CA but different `verify_identity` compare equal and the compat check passes. The case I'm wondering about: connection A sets `verify_identity=False` (writes `SSL_SERVER_DN_MATCH=FALSE`), then a later connection B uses the secure default `verify_identity=True` — B would silently connect without the DN match it asked for, with no error raised. That seems like the same class of silent-reuse the fingerprint was added to prevent for client certs. Should `verify_identity` (and maybe `skip_cert_verification`/`verify_cert`) be folded into `material` here so a differing value fails loudly like a differing CA does? Or is that intentionally out of scope for the single-config-per-thick-agent model?

apollo/integrations/db/oracle_client_config.py
```diff
@@ -0,0 +1,498 @@
...
+
+def create_oracle_thick_wallet(ssl_options: SslOptions) -> Optional[str]:
+    """Build an auto-login Oracle wallet (``cwallet.sso``) for thick-mode TLS.
+
+    Adds the CA cert(s) from ``ca_data`` as trusted certs via orapki, and — when
+    ``cert_data``/``key_data`` are present — imports the client identity for mTLS.
+    Returns the wallet directory, or ``None`` when SSL is disabled / no CA data is
+    provided. On failure the partially-built directory is removed.
+    """
+    if ssl_options.disabled or not ssl_options.ca_data:
```
sassmith, line 238: Nit (feel free to ignore): thick mode always validates the CA chain here and has no equivalent of thin's `skip_cert_verification`/`verify_cert` — a caller that sets `skip_cert_verification=True` expecting to bypass validation will silently get full validation instead. The PR description says this is intended ("the CA chain is always validated in thick mode"); a one-line note in the module docstring calling out that thick doesn't honor those flags would save a future debugging session.

sassmith (approved):

apollo/integrations/db/oracle_client_config.py
mrostan: good catch, fixing ...

apollo/integrations/db/oracle_client_config.py
mrostan: yeah good point, will add a note

apollo/integrations/db/oracle_client_config.py
mrostan: Good catch — you're right, and it's a silent *downgrade* (B asks for the stricter default and gets the looser frozen config). Fixed in 97bb785: `verify_identity` is now folded into the fingerprint material, so a differing value fails loud like a differing CA/client cert, plus a regression test (established `verify_identity=False` → later `True` → raises). I left `skip_cert_verification`/`verify_cert` out on purpose — thick mode ignores them (always validates the chain), so a differing value doesn't change the established config and fingerprinting them would fail loud on an input that has no effect (see the nit thread below, now documented).

apollo/integrations/db/oracle_client_config.py
mrostan: Done in 97bb785 — added a note to `create_oracle_thick_wallet`'s docstring: thick mode always validates the CA chain and has no equivalent of thin's `skip_cert_verification`/`verify_cert`; only `verify_identity` is honored (as `SSL_SERVER_DN_MATCH`).

PR #353, opened 2026-07-22
Add SnowflakeProxyClient.execute_rest_request for agent-routed Snowflake REST (AO-812)

yijiaz-mc:
## What

Adds `SnowflakeProxyClient.execute_rest_request(method, path, body=None, timeout=None) -> dict` so the agent can execute a generic authenticated **JSON** REST call against the customer's Snowflake account, reusing the live `snowflake.connector` session it already holds.

This is the agent-side half of the "generic Snowflake REST via the data collector" feature (AO-812). It lets **agent-based** Snowflake connections invoke the Snowflake REST API (motivating use case: Cortex Agents for golden-set evals) — the data collector can't do this itself for agent connections because it has no local session token; the agent does.

## How

Mirrors the existing `execute_sql_query` and the `SalesforceDataCloudProxyClient.ssot_get` precedent:

- lifts `self._connection.rest.token` + `.server_url` from the live session (guards the pre-login/OAuth `None` case);
- validates `path` is relative (rejects scheme / host / leading `//` / CR-LF) and builds the URL only from the session's own `server_url` — the caller can never redirect it to another host;
- issues the call via `HttpProxyClient.do_request` (default SSRF tier allows private IPs, so PrivateLink works);
- returns `{status_code, response}` on 2xx, `{status_code, error}` (token-redacted) otherwise. The session token is never returned, logged, or echoed in an error.

**JSON only** for now — SSE (streaming) over the agent path is a later follow-up (this transport returns one consolidated result, not a live stream).

The data collector calls this via a reflective `execute_rest_request` command (data-collector PR [#2513](https://github.com/monte-carlo-data/data-collector/pull/2513)).

## Testing

- New unit tests in `tests/test_snowflake_client.py`: JSON success, non-2xx → error, token-`None` guard, unsafe-path rejection. Full file green.

## Context

- [AO-812](https://linear.app/montecarloai/issue/AO-812/data-collector-generic-snowflake-rest-api-job-type-cortex-agent)
- DC side: [data-collector#2513](https://github.com/monte-carlo-data/data-collector/pull/2513)
- Design: SDD "Generic Snowflake REST API requests via DC" (Agent-based connections section)

- [x] Ran `/code-review`

### DIFF
--- a/apollo/integrations/db/salesforce_data_cloud_proxy_client.py
+++ b/apollo/integrations/db/salesforce_data_cloud_proxy_client.py
@@ -15,6 +15,7 @@ from salesforcecdpconnector.exceptions import Error as SalesforceCDPError
 from salesforcecdpconnector.genie_table import GenieTable, Field
 
 from apollo.integrations.db.base_db_proxy_client import BaseDbProxyClient
+from apollo.integrations.http.relative_path import validate_relative_rest_path
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


yijiaz-mc:
Live e2e verified on dev (2026-07-23): a REMOTE_V2 collector dispatched `execute_rest_request` to an agent running pre-release build 3182 (built from this branch via dev), which lifted its session token and executed both a plain REST call and a Cortex agent `:run` against Snowflake successfully. Marking ready for review.

yijiaz-mc:
<details>
<summary>Code Review — 10 findings (7 fixed, 3 skipped)</summary>

**Reviewed:** 2026-07-26 UTC | **Reviewers:** security, correctness, architecture, testing, devdocs | **Scope:** full PR #353

| F1 | ISSUE | Token redaction/body handling applied only on the error branch; docstring promised blanket guarantees the 2xx path didn't keep | ✅ Redact uniformly across all response paths, docstring rewritten to match actual contract |
| F2 | SUGGESTION | Relative-path validation duplicated across proxy clients and already diverging | ✅ Extracted shared `relative_path.py` helper, backported CRLF/protocol-relative checks to `ssot_get` |
| F3 | SUGGESTION | No local timeout default/ceiling — `timeout=None` blocks the agent worker indefinitely | ⏭️ Skipped — DC-side caller pins timeout upstream |
| F4 | SUGGESTION | No size cap or streaming on response body read into agent memory | ⏭️ Skipped — DC-side already caps at 20MB |
| F5 | SUGGESTION | Redaction test was tautological — mocked body never contained the token | ✅ Token embedded in mocked body, test now load-bearing |
| F6 | SUGGESTION | Unsafe-path test didn't mock `requests.request` — a validation regression would trigger real outbound calls in CI | ✅ Mocked `requests.request`, asserts not called on bad paths |
| F7 | SUGGESTION | No test for the 2xx non-JSON fallback path | ✅ Added coverage for the fallback branch |
| F8 | SUGGESTION | Success test never asserted `body`/`timeout` kwargs reached `requests.request` | ✅ Asserts `json`/`timeout` kwargs on `call_args` |
| F9 | SUGGESTION | Plain-relative-path-without-leading-slash branch of validation untested | ✅ Added missing-leading-slash case |
| F10 | SUGGESTION | "No session token" guard only tested `token=None`; `server_url` falsy with token present uncovered | ⏭️ Skipped — low-value edge case, not addressed |

👉 [Fix commit](https://github.com/monte-carlo-data/apollo-agent/commit/4ded287f145788c3c22ca4a5d5469ac408e7da08)

</details>

genehynson (approved):

PR #357, opened 2026-07-27
Add streaming OTLP file writer module (AO-929)

toffer:
## Summary

Adds `apollo.integrations.otlp.otlp_file_writer` — a pure library module implementing [AO-914](https://linear.app/montecarloai/issue/AO-914) contract C4 for the platform-agent trace export effort:

```
write_otlp_files(fragments, storage, key_template, flush_bytes=16MiB, transform=None) -> Manifest
```

It merges a lazy stream of OTLP-JSON fragments (one `{"resourceSpans": [...]}` document per source row) into uncompressed `ExportTraceServiceRequest` files in object storage via `BaseStorageClient`, returning a manifest (`files`, `span_count`, `fragment_count`, `max_collected_ts`, `bytes_written`).

Ticket: [AO-929](https://linear.app/montecarloai/issue/AO-929). Blocks the data-collector `s3-otlp` writer ([AO-934](https://linear.app/montecarloai/issue/AO-934)), which imports this module in-process after the DC pin bump; a later agent operation (M4) wraps the same module for strict-residency accounts. The module has no job/output-stream/dispatch knowledge by design.

## Behavior

- **Batching:** pre-add flush check at `flush_bytes` — no output file exceeds `max(flush_bytes, largest_single_fragment)`. A single oversized fragment ships alone, never truncated (17.44 MiB spans observed in production). Residual buffer flushes at end of stream; empty buffers never produce a file.
- **Fail-fast error contract:** transform, JSON-parse, and missing-`resourceSpans` failures raise `OtlpFragmentError` carrying the 0-based fragment index and the keys already written (at-least-once semantics; deterministic key templates make retries overwrite cleanly). Errors never carry fragment content — fragments may hold prompt/completion text, and exceptions flow to logs/Sentry. Parse errors are deliberately unchained (`JSONDecodeError.doc` retains the full document); transform errors are referenced by type only.
- **Keys:** module fills `key_template.format(seq=N)`, 0-based, canonical `{seq:04d}` per the C1 key spec. The template is relative to the storage client's namespace; `Manifest.files` (and `keys_written`) report keys as passed.
- **Memory:** peak is a small multiple of `flush_bytes` and unbounded by a single oversized fragment (documented honestly for Lambda callers); fragments are consumed lazily — the stream is never materialized.

## Testing

18 unit tests (`tests/test_otlp_file_writer.py`) against a mocked `BaseStorageClient` with real-shaped OTLP-JSON fixtures (`resourceSpans → scopeSpans → spans`, attribute key/value lists): merge shape and order, pre-add flush bound, UTF-8 byte-vs-char boundary, oversized fragment, final flush, empty stream, all three failure classes incl. failure-after-flush, error content hygiene, manifest fields verified against actual mock payloads, lazy consumption, and canonical seq numbering.

Verified: `black --check`, `pyright` (0 errors), `pytest` 18/18; wheel build confirms `apollo.integrations.otlp` ships in the package and imports from a clean install.

## Key Decisions

- **Pre-add flush check** (files bounded at ~flush_bytes) over post-add (~33 MiB worst case) — pinned in the AO-914 C1 contract.
- **Content-free exceptions** are contract, not style: `keys_written` gives the caller orphan-cleanup control without ever exposing payload text.
- **Output is raw UTF-8** (`ensure_ascii=False`, compact separators) — smaller files, standard for OTLP-JSON.
- **Not adopted** (recorded in the plan): raw-string buffering (halves steady-state memory, doubles parse CPU) and a `max_fragment_bytes` cap — the documented realistic memory bound covers the sizing risk for v1.
- **No PII filtering here**: content is written as-is; if AO-931's open `pii_filters` decision lands on this path, the `transform` hook is the application point.

## Release

After merge: GitHub Release **v1.10.0** (latest is v1.9.10) so AO-934 can bump the DC pin from v1.7.2. Note the Release also publishes agent Docker images to DockerHub/ECR — full-release timing to be confirmed before cutting (tag-only fallback documented in the plan).

---

- [x] Ran `/code-review`

### DIFF
--- a/apollo/integrations/otlp/otlp_file_writer.py
+++ b/apollo/integrations/otlp/otlp_file_writer.py
@@ -39,10 +39,12 @@ Contract notes (see the AO-914 master plan, contracts C1/C4):
   with the same deterministic key template overwrites them. Errors never carry
   fragment content — fragments may hold prompt/completion text and exceptions
   flow to logs and error reporting. Transform exceptions are referenced by
-  type only and are not chained (``raise ... from None``), so a transform's
-  own exception message — which may embed fragment content — never reaches
-  ``__cause__``; the JSON-parse path applies the same treatment for the same
-  reason.
+  type only, and the underlying exception is dropped inside the ``except``
+  handler so :class:`OtlpFragmentError` is raised outside it: this leaves both
+  ``__cause__`` and the implicit ``__context__`` as ``None``, so a transform's
+  own exception message — which may embed fragment content — is unreachable
+  through the exception chain. The JSON-parse path applies the same treatment
+  for the same reason (``JSONDecodeError.doc`` holds the full raw fragment).
 - **Memory:** peak usage is a small multiple of ``flush_bytes`` (parsed object
   trees plus the merged output string at flush time) and is unbounded by a
   single oversized fragment. Fragments are consumed lazily from the iterator;
@@ -150,19 +152,22 @@ def write_otlp_files(
 
     for index, raw_fragment in enumerate(fragments):
         if transform is not None:
+            transform_error: Optional[str] = None
             try:
                 raw_fragment = transform(raw_fragment)
             except Exception as exc:
-                # Deliberately not chained: a transform's own exception
-                # message may embed fragment content (e.g. prompt/completion
-                # text). The exception type is already captured in the detail
-                # string, so no diagnostic value is lost by breaking the
-                # chain.
-                raise OtlpFragmentError(
-                    index,
-                    f"transform raised {type(exc).__name__}",
-                    list(files),
-                ) from None
+                # Capture a content-free detail only. A transform's own
+                # exception message may embed fragment content (e.g.
+                # prompt/completion text), so the exception object itself must
+                # not survive this handler.
+                transform_error = f"transform raised {type(exc).__name__}"
+            if transform_error is not None:
+                # Raised outside the ``except`` block so Python leaves
+                # ``__context__`` as ``None``. ``from None`` alone would clear
+                # ``__cause__`` and set ``__suppress_context__``, but the
+                # implicit ``__context__`` chain would still reference the
+                # transform's exception — and thus its fragment-bearing message.
+                raise OtlpFragmentError(index, transform_error, list(files))
 
         if not isinstance(raw_fragment, str):
             raise OtlpFragmentError(
@@ -171,16 +176,21 @@ def write_otlp_files(
                 list(files),
             )
 
+        parse_error: Optional[str] = None
         try:
             document = json.loads(raw_fragment)
         except json.JSONDecodeError as exc:
-            # Deliberately not chained: JSONDecodeError retains the full
-            # document on its .doc attribute, which may hold prompt content.
-            raise OtlpFragmentError(
-                index,
-                f"invalid JSON: {exc.msg} at line {exc.lineno} column {exc.colno}",
-                list(files),
-            ) from None
+            # Capture position/message only. JSONDecodeError retains the full
+            # document on its ``.doc`` attribute (may hold prompt content), so
+            # the exception object itself must not survive this handler.
+            parse_error = (
+                f"invalid JSON: {exc.msg} at line {exc.lineno} column {exc.colno}"
+            )
+        if parse_error is not None:
+            # Raised outside the ``except`` block so ``__context__`` stays
+            # ``None`` and the fragment-bearing ``JSONDecodeError`` is not
+            # reachable through the implicit exception chain.
+            raise OtlpFragmentError(index, parse_error, list(files))
 
         resource_spans = (
             document.get(_RESOURCE_SPANS_KEY) if isinstance(document, dict) else None
--- a/apollo/integrations/otlp/otlp_file_writer.py
+++ b/apollo/integrations/otlp/otlp_file_writer.py
@@ -1,0 +1,233 @@
+"""Streaming OTLP-JSON file writer (AO-914 contract C4).
+
+Turns a stream of OTLP-JSON fragments — each a self-contained
+``{"resourceSpans": [...]}`` document, one per source row — into uncompressed
+OTLP-JSON files in object storage. Each output object is exactly one
+``ExportTraceServiceRequest``-shaped document whose ``resourceSpans`` list is
+the concatenation of the buffered fragments' lists.
+
+This is a pure library module: it has no knowledge of jobs, output streams, or
+dispatch modes. Callers (the data collector in-process; later a remote agent
+operation) supply an already-authenticated storage client and a key template.
+
+Contract notes (see the AO-914 master plan, contracts C1/C4):
+
+- **Batching:** a file is flushed when adding the next fragment would cross
+  ``flush_bytes`` (pre-add check), so no output file exceeds
+  ``max(flush_bytes, largest_single_fragment)``. A single fragment larger than
+  the threshold ships alone, never truncated. The residual buffer is flushed
+  when the iterator is exhausted; an empty buffer never produces a file.
+- **Keys:** ``key_template`` is filled via ``key_template.format(seq=N)`` with
+  ``seq`` 0-based and incremented once per file; the canonical template uses
+  ``{seq:04d}`` (pinned in C1) but the module is format-agnostic. The template
+  is relative to the storage client's namespace — ``BaseStorageClient``
+  prepends its configured prefix — and ``Manifest.files`` reports keys as
+  passed (pre-prefix); aligning client prefix + template with the ingest path
+  is the caller's responsibility.
+- **Failure semantics:** any fragment-level failure (transform error, JSON
+  parse error, missing/invalid ``resourceSpans``) raises
+  :class:`OtlpFragmentError` and aborts the run. Semantics are at-least-once:
+  files flushed before the failure remain in storage (``keys_written`` on the
+  error lists them, same as-passed namespace as ``Manifest.files``); a retry
+  with the same deterministic key template overwrites them. Errors never carry
+  fragment content — fragments may hold prompt/completion text and exceptions
+  flow to logs and error reporting. Wrapped transform exceptions are
+  referenced by type only; the chained ``__cause__`` of a transform failure is
+  the caller's content-hygiene responsibility.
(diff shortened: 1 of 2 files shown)


toffer:
<details>
<summary>Code Review — 7 findings (6 fixed, 1 skipped)</summary>

**Reviewed:** 2026-07-27 UTC | **Reviewers:** security, devdocs, architecture, correctness, testing | **Scope:** full PR

| F1 | BLOCKER | Transform-error path chained the original exception (`from exc`), risking a PII-bearing exception reaching `__cause__` and getting logged | ✅ Changed to `from None`; test asserts `__cause__ is None` |
| F2 | ISSUE | Non-str fragment (e.g. `transform` returning `None`) raised a bare `TypeError` that escaped the documented `OtlpFragmentError` envelope | ✅ Added `isinstance` guard raising `OtlpFragmentError` with content-free detail |
| F3 | SUGGESTION | Batching docstring documented only the pre-add flush check, omitting the load-bearing post-add check | ✅ Docstring + param doc now describe both triggers |
| F4 | SUGGESTION | Documented "hard" size bound can be exceeded by a few bytes due to JSON number re-serialization | ✅ Docstring softened to present the bound as approximate |
| F5 | SUGGESTION | `max_collected_ts` "max" semantics never distinguished from "last-seen wins" in tests | ✅ Added non-monotonic cross-fragment and intra-fragment max tests |
| F6 | SUGGESTION | Best-effort defensive parsing in `_collect_span_stats`/`_list_of_dicts` had zero coverage | ✅ Added malformed-nested-levels graceful-skip test |
| F7 | SUGGESTION | Three minor consolidated test-coverage gaps (non-dict top-level doc, multiple resourceSpans/scopeSpans per fragment, `max_collected_ts` None on non-empty run) | ⏭️ Skipped — low confidence |

👉 [Fix commit](https://github.com/monte-carlo-data/apollo-agent/commit/0e24069)

</details>

apollo/integrations/otlp/otlp_file_writer.py
```diff
@@ -0,0 +1,256 @@
...
+        if transform is not None:
+            try:
+                raw_fragment = transform(raw_fragment)
+            except Exception as exc:
+                # Deliberately not chained: a transform's own exception
+                # message may embed fragment content (e.g. prompt/completion
+                # text). The exception type is already captured in the detail
+                # string, so no diagnostic value is lost by breaking the
+                # chain.
+                raise OtlpFragmentError(
```
ncastagnet: From Claude: The content-hygiene contract has a gap on the exception `__context__` chain. `raise OtlpFragmentError(...) from None` clears `__cause__` and sets `__suppress_context__=True`, but because both raises execute inside the `except` block, Python still sets `__context__` to the original exception — on the transform path that's the transform's own exception (its message may embed fragment content -> `str(err.__context__)` leaks it); on the JSON path it's the `JSONDecodeError`, whose `.doc` holds the full raw fragment. Standard channels (traceback/logging/Sentry) honor `__suppress_context__`, so this only bites a reporter that walks `__context__` directly — but that's the leak the "errors never carry fragment content" contract exists to prevent. The two hygiene tests only assert `__cause__ is None` + `repr(vars(exc))`, neither of which touches `__context__`, so they'd stay green even with the leak. Note `err.__context__ = None; raise err` does NOT work (raising inside the `except` re-sets it); the fix is to capture the content-free detail in the handler and `raise` after the `except` block. Worth adding `assertIsNone(exc.__context__)` (fails today) + a `traceback.format_exception` full-chain scan to the hygiene tests.

ncastagnet (approved):
Looks good!

apollo/integrations/otlp/otlp_file_writer.py
toffer: Fixed in 7a28549. Both `OtlpFragmentError` raises now happen *outside* their `except` blocks — each handler captures a content-free detail string and drops the underlying exception, so Python leaves both `__cause__` and `__context__` as `None`. Confirmed the transform message and `JSONDecodeError.doc` are no longer reachable through the exception chain.

Extended both hygiene tests to assert `__context__ is None` and to scan the full `traceback.format_exception` chain for the secret. Test-first: both new assertions failed against the old code, pass now (22/22).

toffer:
## PR Feedback — 1 item addressed

**Pushed:** 7a28549

| # | Item | Status |
|---|------|--------|
| H1 | Exception `__context__` chain leaks fragment content despite the content-free contract ([thread](https://github.com/monte-carlo-data/apollo-agent/pull/357#discussion_r3668750814)) | ✅ Fixed |

**H1** — Both `OtlpFragmentError` raises ran inside their `except` blocks, so despite `from None` Python still populated `__context__` with the original exception (the transform's own message on the transform path; `JSONDecodeError.doc`, the full raw fragment, on the parse path). Now each handler captures a content-free detail string and drops the underlying exception, and the raise fires *after* the block — leaving both `__cause__` and `__context__` as `None`.

Both hygiene tests were extended to assert `__context__ is None` and to scan the full `traceback.format_exception` chain for the secret. Test-first confirmed the new assertions fail against the old code; all 22 tests pass after the fix. `black` + `pyright` clean.

mrostan:
## Review — 5 findings, none blocking

Read the full module, tests, `BaseStorageClient`, and packaging. Good PR — pure library boundary is clean, lazy consumption is real, fixtures are real-shaped, CI green. The `__context__` fix in `3df2fb7` is genuinely correct and easy to get wrong: `from None` alone would have left `JSONDecodeError.doc` reachable through the implicit chain. Findings worst-first.

### 1. `storage.write` failure loses `keys_written` — contract asymmetry
`otlp_file_writer.py:147`

Fragment-level errors hand the caller `keys_written` for orphan cleanup. A mid-run storage failure (S3 throttle, transient 5xx) raises `BaseStorageClient.GenericError` bare — no key list — and that's precisely the partial-run case where cleanup matters most. Either wrap the `storage.write` call and re-raise carrying the accumulated keys, or state explicitly in the failure-semantics docstring that storage errors are outside the `OtlpFragmentError` envelope. Currently untested either way.

### 2. Memory bound is optimistic for the Lambda callers
`otlp_file_writer.py:48`

`buffered_bytes` counts raw UTF-8 fragment bytes, but `buffer` holds parsed object trees. OTLP-JSON is dict/str-dense, so 5–15× raw is typical. At the default 16 MiB threshold that's roughly 100–250 MB resident, plus the merged `str` and its `.encode()` copy at flush time. "A small multiple of `flush_bytes`" undersells that for AO-934 (DC in-process) and M4 sizing. Worth measuring once against production-shaped fragments and documenting the real multiplier.

Free win in the same place (`:149`): `json.dumps(...)` followed by `payload.encode("utf-8")` holds both copies at peak. `BaseStorageClient.write` accepts `Union[bytes, str]` — encode once, write the bytes, `bytes_written += len(payload)`. Drops a full-payload copy at the worst moment.

### 3. No `{seq}` validation → silent overwrite
`otlp_file_writer.py:146`

`key_template.format(seq=N)` against a template with no `{seq}` placeholder produces the same key on every flush: each file silently overwrites the previous one, `Manifest.files` lists the same key N times, and the run reports success. A stray `{`/`}` in the template raises a bare `KeyError`/`ValueError` outside the documented error envelope. A one-line guard at entry (`if "{seq" not in key_template: raise ValueError(...)`) closes both. No test covers either case.

### 4. `Iterator[str]` should be `Iterable[str]`
`otlp_file_writer.py:105`

`enumerate` needs only an iterable. As typed, a caller passing a `list` gets a pyright error for no reason. Laziness is unaffected by the widening.

### 5. Doc nit — post-add flush isn't a co-equal trigger
`otlp_file_writer.py:17`

Pre-add fires whenever `buffered + next > flush_bytes`, so the post-add check (`buffered >= flush_bytes`) can only fire on exact-boundary equality, or on an empty buffer holding a single oversized fragment. In the oversized case the resulting file is identical either way — the next fragment's pre-add check, or the final flush, produces the same content. The post-F3 docstring reads as two equally load-bearing triggers; it's really one trigger plus an edge case. Behavior is fine, only the wording overstates.

mrostan (approved):
approved, left a few comments by Claude

toffer:
@mrostan Thanks for the review — all 5 findings addressed in 14fb60c. Two were fixed with a different remedy than proposed (H1, H3), rationale below.

**H1 — storage.write failure loses `keys_written`** ✅ Fixed, different shape than suggested. Wrapping in `OtlpFragmentError` doesn't quite work: the terminal flush runs after the fragment loop, so there's no meaningful `fragment_index` at that point — and it would conflate retryable infra faults with permanent data faults. Instead there's a new `OtlpStorageError(key, keys_written)` carrying the failed key plus all keys already written, chained `from` the backend exception (storage errors carry no fragment content, so the chain is PII-safe). Failure-semantics docstring updated; test added asserting `keys_written` and `__cause__` survive.

**H2 — memory docs optimistic; encode-once** ✅ Both taken. The docstring now states a shape-dependent multiplier range measured on OTLP-shaped fragments: ~5.9–10.8x raw bytes for attribute-dense payloads (~170 MB resident at the 16 MiB default), ~1.2–2.7x for prompt/completion-heavy ones. `flush()` now encodes once and writes bytes — verified behaviorally identical across S3/GCS/Azure backends; `bytes_written` unchanged.

**H3 — no `{seq}` validation** ✅ Fixed, stronger guard than suggested. The substring check (`"{seq" not in key_template`) misses stray-brace templates like `{seq:04d}}` or `{seq}_{oops}`, which still blow up mid-run at first flush. Instead the template is probe-formatted at entry (`format(seq=0)` vs `seq=1`, must differ; format errors caught) and raises `ValueError` before the fragment iterator is touched — so bad templates fail fast even on an empty stream. `:raises ValueError:` documented; tests added for missing-placeholder, constant-key, and malformed templates.

**H4 — `Iterator[str]` → `Iterable[str]`** ✅ Done, plus removed the now-unneeded `iter(...)` wrapping at test call sites.

**H5 — post-add flush doc wording** ✅ Reworded: pre-add is the file-boundary trigger; post-add is documented as an output-redundant early-release detail (exact boundary / single oversized fragment). Simulation confirmed post-add never changes output, only buffer-release timing.

Verification: `black` clean, `pyright` 0 errors, 25/25 tests passing (3 new).

PR #364, opened 2026-08-03
fix(ctp): route credential-resolution requests through SSRF-guarded safe_request

mrostan:
## Summary

Four CTP credential-resolution transforms build HTTP requests to **customer-supplied, template-rendered endpoints** and call `requests.post` directly, bypassing the SSRF guard:

- `oauth.py` — OAuth `token_endpoint` (C4)
- `resolve_informatica_session.py` — Informatica V2 login (C5), V3 login (C6), JWT/OAuth login (C7)

`apollo/integrations/http/CLAUDE.md` states that request paths not going through `HttpProxyClient.do_request` **must** wrap with `safe_request`. These four don't — so this routes them through it.

## Change

`requests.post(url, …)` → `safe_request("POST", url, …)` (a documented drop-in for `requests.request`). **Default tier**: blocks cloud-metadata / loopback destinations, while still allowing legitimate **private/on-prem RFC1918** IdP and Informatica hosts — so existing customer connections keep working. No behavior change beyond the destination-validation guard.

## Test updates

The transforms now issue requests via `safe_request` instead of `requests.post`, so the tests that mocked the old call path were updated (8 files, ~60 tests):

- Patch the module-level `safe_request` name (instead of `requests.post` / `oauth.requests`) so mocks intercept the new call and validation/DNS is fully bypassed — pure unit isolation.
- `safe_request(method, url, **kwargs)` puts the URL at the **second** positional arg, so URL assertions moved from `[0][0]` → `[0][1]`.
- `test_informatica_v2_ctp`: the v2 pipeline issues its two HTTP calls from two different transforms — a small helper patches both to one shared mock so the ordered first-call/second-call assertions keep working.
- Dropped now-unneeded requests-module exception-attr restores in the mulesoft failure test.

Full suite green locally (1450 passed).

## Provenance

Extracted from the **code-only** portion of rybot's SAST PR #361 (findings C4–C7, each Tank-triaged REAL_VULN). The AI-tooling files rybot also staged there (`.claude/skills/fix-repo-vulns/*`, `.mcp.json`) are **intentionally excluded** — they don't belong in this public repo. (Note: rybot's #361 changed the code without updating these tests, so it would fail CI as-is — another reason to take this PR instead.)

## Checklist
- [x] Code changes verified locally (safe_request is a drop-in; default tier preserves RFC1918)
- [x] Aligned with repo SSRF policy (`apollo/integrations/http/CLAUDE.md`)
- [x] Tests updated for the new call path — full suite green (1450 passed)
- [na] Docs — policy already documented in http/CLAUDE.md

> Note: `oauth.py` is also touched by the in-flight dependency PR (#360) on a different line (its `HTTPError` handler); no conflict expected in either merge order.

### DIFF
--- a/apollo/integrations/ctp/transforms/oauth.py
+++ b/apollo/integrations/ctp/transforms/oauth.py
@@ -2,6 +2,7 @@ import base64
 
 import requests
 
+from apollo.integrations.http.url_safety import safe_request
 from apollo.integrations.ctp.errors import CtpPipelineError
 from apollo.integrations.ctp.models import PipelineState, TransformStep
 from apollo.integrations.ctp.template import TemplateEngine
@@ -97,8 +98,12 @@ class OAuthTransform(Transform):
         }
 
         try:
-            response = requests.post(
-                token_endpoint, data=data, headers=headers, timeout=30
+            # SSRF guard: token_endpoint comes from customer connection config
+            # (template-rendered), so route through the SSRF-guarded request
+            # helper, which blocks cloud-metadata / loopback destinations while
+            # still allowing legitimate (incl. private/on-prem) IdP hosts.
+            response = safe_request(
+                "POST", token_endpoint, data=data, headers=headers, timeout=30
             )
             response.raise_for_status()
         except requests.HTTPError as exc:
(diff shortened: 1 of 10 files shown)


walles2 (approved):
lgtm - so, whenever we're using a request with credentials, we should use `safe_request`

